PR #13, opened 2026-04-20
Update bootstrap support

nyagamunene:
Pull request title should be `NOISSUE - description` or `MG-XXX - description` where XXX is the ID of the issue this PR relates to.
Please review the [CONTRIBUTING.md](https://github.com/absmach/magistrala/blob/main/CONTRIBUTING.md) file for detailed contributing guidelines.

Put here `Resolves #XXX` to auto-close the issue that your PR fixes (if such)

### Notes


pkg/nodered/client.go
```diff
@@ -161,14 +179,76 @@ func normalizeAddFlowPayload(flow string) ([]byte, error) {
...
+			collectNodeRedIDs(child, ids)
+		}
+	case []any:
+		for _, child := range v {
+			collectNodeRedIDs(child, ids)
+		}
+	}
+}
+
+func rewriteNodeRedIDs(value any, ids map[string]string) {
```
rodneyosodo, line 226: **`rewriteNodeRedIDs` replaces every string value that matches a collected ID, not just fields that semantically represent ID references** (e.g. `id`, `broker`, `z`, `tls`, `wires`). If a flow has a `name`, `label`, or `info` field whose value coincidentally equals another node's `id`, it will be silently rewritten to a random `nr-xxx` ID.

This is low severity because Node-RED auto-generated IDs are long hex strings, but example flows in this repo use human-readable IDs like `mqtt-broker-config` and `flow-speed-sensor` which are more collision-prone.

**Suggestion:** Maintain an explicit set of ID-reference fields and only rewrite those:

```go
var idFields = map[string]bool{
    "id": true, "broker": true, "z": true, "tls": true,
}

func rewriteNodeRedIDs(value any, ids map[string]string) {
    switch v := value.(type) {
    case map[string]any:
        for key, child := range v {
            if typed, ok := child.(string); ok && idFields[key] {
                if replacement, ok := ids[typed]; ok {
                    v[key] = replacement
                }
            } else {
                rewriteNodeRedIDs(typed, ids)
            }
        }
    case []any:
        for i, child := range v {
            switch typed := child.(type) {
            case string:
                if replacement, ok := ids[typed]; ok {
                    v[i] = replacement
                }
            default:
                rewriteNodeRedIDs(typed, ids)
            }
        }
    }
}
```

Note: this would also need to handle `wires` arrays (which contain arrays of ID strings) specially.

pkg/bootstrap/bootstrap.go
```diff
@@ -212,7 +214,11 @@ func getConfig(bsID, bsKey, bsSvrURL string, skipTLS bool, logger *slog.Logger)
 		return deviceConfig{}, err
 	}
 
-	req.Header.Add("Authorization", fmt.Sprintf("Client %s", bsKey))
+	authScheme := "Client"
+	if strings.Contains(bsSvrURL, "/things/bootstrap") {
```
rodneyosodo: **`strings.Contains` on the full URL is fragile** — it matches the substring `/things/bootstrap` anywhere in the URL, including in the hostname or query parameters (e.g. `https://things-bootstrap.example.com/api` would incorrectly match).

**Suggestion:** Parse the URL and inspect only the path component:

```go
authScheme := "Client"
parsed, err := url.Parse(bsSvrURL)
if err == nil && strings.HasPrefix(parsed.Path, "/things/bootstrap") {
    authScheme = "Thing"
}
```

The `url` import is already available in this package.

pkg/bootstrap/bootstrap.go
```diff
@@ -112,7 +113,7 @@ func Bootstrap(cfg Config, logger *slog.Logger, file string) error {
 		}
 	}
 
-	if len(dc.Channels) < 2 {
+	if len(dc.Channels) < 1 {
```
rodneyosodo: **Behavioral change:** The minimum channel count validation was lowered from 2 to 1. This means bootstrap configurations with only a single channel are now accepted, which could affect existing setups that relied on the 2-channel minimum as a validation guard.

If this is intentional (e.g. the agent now operates with a single channel), consider adding a brief note in the PR description to make this change explicit.

scripts/provision.sh
```diff
@@ -0,0 +1,503 @@
...
+echo "Domain ID:    ${DOMAIN_ID}"
+echo ""
+
+TOKEN="${MG_PAT}"
+if [ -z "$TOKEN" ]; then
+  echo "ERROR: MG_PAT environment variable is not set."
+  exit 1
+fi
+echo "Step 1: Using PAT token (${#TOKEN} characters)."
+echo "Token starts with: ${TOKEN:0:30}"
```
rodneyosodo: **Partial token content is logged to stdout.** Even token fragments can be useful to an attacker, especially combined with the known token format (`pat_` prefix, base64 structure).

```bash
echo "Token starts with: ${TOKEN:0:30}"
echo "Token ends with: ${TOKEN: -20}"
```

And on line 293:
```bash
echo "  Token: ${TOKEN:0:20}..." # Show first 20 chars only
```

**Suggestion:** Gate these behind a verbose/debug flag or remove them:

```bash
if [ "${VERBOSE:-}" = "true" ]; then
    echo "Token starts with: ${TOKEN:0:8}..."
fi
```

Or simply log the token length without content: `echo "Step 1: Using PAT token (${#TOKEN} characters)."`

scripts/entrypoint.sh
```diff
@@ -0,0 +1,175 @@
...
+export MQTT_USETLS
+export MQTT_SKIP_TLS
+
+if [ ! -f /data/.initialized ]; then
+    cp /seed/nodered/settings.js /data/settings.js
+    cp /seed/nodered/flows.json  /data/flows.json
+    touch /data/.initialized
+fi
+
+cat > /data/flows_cred.json << EOF
```
rodneyosodo: **This `cat > /data/flows_cred.json` block is dead code.** It writes a hardcoded `mqtt-broker-config` credential entry, but the Node.js section starting at line ~140 overwrites the same file with credentials keyed by the actual broker IDs from the flow.

The initial write on this line is never read — it's immediately overwritten by `fs.writeFileSync(credFile, ...)` at line 172.

**Suggestion:** Remove lines 85-92 to avoid confusion.

rodneyosodo (commented):
## Review Summary

Overall the changes are solid. Found a few issues worth addressing:

1. **`rewriteNodeRedIDs` replaces all matching strings**, not just ID-reference fields — could corrupt non-ID fields that happen to match a node ID
2. **Auth scheme detection uses `strings.Contains` on full URL** — fragile, could match in hostname/query params
3. **Minimum channel count changed from 2→1** — behavioral change not explicitly called out in PR description
4. **`provision.sh` logs partial token content** — token fragments can aid attackers
5. **Dead code in `entrypoint.sh`** — initial `flows_cred.json` write is immediately overwritten

cmd/main.go
```diff
@@ -131,13 +168,44 @@ func main() {
...
+	}
+	if len(missing) > 0 {
+		return errors.New(fmt.Sprintf("%s: missing required runtime fields: %s", errInvalidRuntimeConfig, strings.Join(missing, ", ")))
+	}
+	return nil
+}
+
+func hasBootstrapConfig(cfg config) bool {
+	return cfg.BootstrapURL != "" || cfg.BootstrapID != "" || cfg.BootstrapKey != ""
+}
```
rodneyosodo, line 200: **Bug: OR vs AND mismatch with `loadBootConfig`**

`hasBootstrapConfig` returns `true` if *any one* of the three vars is set (OR), but `loadBootConfig` (called right after) requires *all three* to be set (AND). If a user accidentally sets only `MG_AGENT_BOOTSTRAP_ID` without the other two, the agent enters bootstrap mode and immediately fails with a confusing error message.

Change to AND so the guard matches the requirement:
```go
func hasBootstrapConfig(cfg config) bool {
    return cfg.BootstrapURL != "" && cfg.BootstrapID != "" && cfg.BootstrapKey != ""
}
```

rodneyosodo (commented):
Good overall direction — bootstrap mode is well-architected and the package restructure is clean. Found a few concrete bugs and one doc issue worth fixing before merge. See inline comments.

cmd/main.go
```diff
@@ -131,13 +168,44 @@ func main() {
...
+	if cfg.DomainID == "" {
+		missing = append(missing, "domain_id")
+	}
+	if cfg.Channels.CtrlChan() == "" || cfg.Channels.DataChan() == "" {
+		missing = append(missing, "channels.id (or channels.ctrl_id + channels.data_id)")
+	}
+	if cfg.MQTT.URL == "" {
+		missing = append(missing, "mqtt.url")
+	}
+	if cfg.MQTT.Username == "" {
```
rodneyosodo: **Bug: `validateRuntimeConfig` rejects valid mTLS setups**

Requiring both `mqtt.username` and `mqtt.password` breaks certificate-only mTLS auth where credentials are replaced by a client cert. Consider skipping these checks when `cfg.MQTT.MTLS` is `true`:

```go
if !cfg.MQTT.MTLS {
    if cfg.MQTT.Username == "" {
        missing = append(missing, "mqtt.username")
    }
    if cfg.MQTT.Password == "" {
        missing = append(missing, "mqtt.password")
    }
}
```

cmd/main.go
```diff
@@ -191,20 +257,80 @@ func loadEnvConfig(cfg config) (agent.Config, error) {
...
+	if c.MQTT.CAPath == "" {
+		c.MQTT.CAPath = defaults.MQTT.CAPath
+	}
+	if c.MQTT.CertPath == "" {
+		c.MQTT.CertPath = defaults.MQTT.CertPath
+	}
+	if c.MQTT.PrivKeyPath == "" {
+		c.MQTT.PrivKeyPath = defaults.MQTT.PrivKeyPath
+	}
+	if c.MQTT.QoS == 0 {
```
rodneyosodo: **Bug: QoS 0 is a valid config value, not a zero-value sentinel**

This overwrites a config-file QoS of 0 with the env default (which is also 0, so it's harmless today, but semantically wrong). QoS 0 is a legitimate and common setting — using it as a "not set" sentinel is incorrect.

Options:
- Use a pointer `*byte` in `MQTTConfig` to distinguish "unset" from 0.
- Or remove this merge entry entirely — if both default to 0, nothing needs merging.

api/endpoints_test.go
```diff
@@ -0,0 +1,559 @@
...
+
+			if tc.status != http.StatusOK {
+				return
+			}
+
+			var body genericBody
+			err = json.NewDecoder(res.Body).Decode(&body)
+			assert.Nil(t, err, fmt.Sprintf("%s: unexpected error while decoding response body: %s", tc.desc, err))
+			assert.Equal(t, "agent", body.Service)
+			assert.Equal(t, "config", body.Response)
```
rodneyosodo: **Bug: wrong assertion — `/pub` response is not `"config"`**

The `/pub` endpoint doesn't return a generic body with `response: "config"`. This assertion will pass vacuously when the body field is empty, meaning the test isn't actually verifying the response content. Same copy-paste issue exists in `TestAddConfig` at line 349.

Check what the publish endpoint actually returns and assert against that, or drop the body check if the response body is intentionally empty on success.

README.md
```diff
@@ -34,24 +34,101 @@ The recommended way to run agent is with the provided Docker Compose stack, whic
...
-MG_AGENT_CHANNEL=<channel-id>
+MG_AGENT_BOOTSTRAP_ID=<external-id>
+MG_AGENT_BOOTSTRAP_KEY=<external-key>
+MG_AGENT_BOOTSTRAP_URL=http://bootstrap:9013/clients/bootstrap
+```
+
+You can fetch the rendered bootstrap response directly:
+
+```bash
+postman request 'http://localhost:9013/clients/bootstrap/01:6:0:sb:sa' \
```
rodneyosodo: **Doc bug: `postman request` is not a CLI command**

`postman` is a GUI application — it has no `request` CLI subcommand available by default. This example will fail for anyone who tries to run it. Replace with `curl`:

```bash
curl 'http://localhost:9013/clients/bootstrap/01:6:0:sb:sa' \
  --header 'accept: */*' \
  --header 'Authorization: Client secret'
```

Makefile
```diff
@@ -137,13 +137,16 @@ release:
 	$(call docker_push,$(version))
 
 run_provision:
-	@MG_API="$(MG_API)" MG_DOMAIN_ID="$(MG_DOMAIN_ID)" MG_PAT="$(MG_PAT)" bash docker/nodered/provision.sh
+	@bash scripts/provision.sh
+
+provision:
```
rodneyosodo, line 141: **Nit: duplicate targets**

`run_provision` (line 139) and `provision` (line 142) both run the same script. Pick one canonical name. If backward-compat with existing `make run_provision` calls matters, alias it:
```makefile
run_provision: provision

provision:
	@bash scripts/provision.sh
```

pkg/bootstrap/bootstrap.go
```diff
@@ -56,189 +30,199 @@ type Config struct {
...
+// FetchAgentConfig retrieves device configuration from the bootstrap service and
+// overlays the returned credentials and channel IDs onto agentCfg.
+func FetchAgentConfig(cfg Config, agentCfg agent.Config, logger *slog.Logger) (agent.Config, error) {
 	retries, err := strconv.ParseUint(cfg.Retries, 10, 64)
 	if err != nil {
-		return errors.New(fmt.Sprintf("Invalid BOOTSTRAP_RETRIES value: %s", err))
+		return agentCfg, errors.New(fmt.Sprintf("Invalid BOOTSTRAP_RETRIES value: %s", err))
 	}
-
 	if retries == 0 {
```
rodneyosodo, line 75: **Nit: silent clamp should be logged**

Clamping `retries == 0` to 1 silently changes the caller's intent. At minimum log a warning so operators know the value was adjusted:
```go
if retries == 0 {
    logger.Warn("BOOTSTRAP_RETRIES is 0, clamped to 1")
    retries = 1
}
```

pkg/bootstrap/bootstrap.go
```diff
@@ -56,189 +30,199 @@ type Config struct {
...
+	if err := json.Unmarshal(body, &br); err != nil {
+		return bootstrapResponse{}, err
 	}
-	dc.SvcsConf = sc
-	return dc, nil
+
+	return br, nil
+}
+
+func bootstrapConfigURL(bsSvrURL, bsID string) string {
```
rodneyosodo, line 226: **Nit: only trims single-slash prefix, not all leading slashes**

`strings.TrimLeft(bsID, "/")` removes *all* leading slashes (not just one) because `TrimLeft` takes a cutset, not a prefix. The test labeled "trim duplicate slashes" confirms this. The function name and comment are fine, but the test description is slightly misleading — worth renaming to "trim all leading slashes from id".

rodneyosodo (approved):

PR #15, opened 2026-05-08
NOISSUE - Update senml encoder

nyagamunene:
Pull request title should be `NOISSUE - description` or `MG-XXX - description` where XXX is the ID of the issue this PR relates to.
Please review the [CONTRIBUTING.md](https://github.com/absmach/magistrala/blob/main/CONTRIBUTING.md) file for detailed contributing guidelines.

Put here `Resolves #XXX` to auto-close the issue that your PR fixes (if such)

### Notes


pkg/conn/conn.go
```diff
@@ -123,19 +123,19 @@ func extractBrokerTopic(topic string) string {
...
-	if len(sm.Records) == 0 {
+	if len(records) == 0 {
 		b.logger.Error("SenML payload empty", slog.Any("payload", msg.Payload()))
 		return
 	}
-	cmdType := sm.Records[0].Name
-	cmdStr := *sm.Records[0].StringValue
-	uuid := strings.TrimSuffix(sm.Records[0].BaseName, ":")
+	cmdType := records[0].Name
+	cmdStr := *records[0].StringValue
```
JeffMboya: `records[0].StringValue` is not guarded: if the payload carries a float or bool record (both now encodable via the new helpers) `StringValue` will be `nil` and this dereference panics. Check for nil before dereferencing.

pkg/senml/senml.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// EncodeBool encodes a single boolean record.
+func EncodeBool(bn, n string, v bool) ([]byte, error) {
+	return pack(bn, n, senml.Record{BoolValue: &v})
+}
+
+// EncodeRecords encodes an arbitrary slice of records as a SenML JSON array.
+func EncodeRecords(records []senml.Record) ([]byte, error) {
+	return encode(senml.Pack{Records: records})
```
JeffMboya, line 52: `EncodeRecords` calls `encode` directly, skipping `pack`, so no timestamp is injected. All other `Encode*` helpers stamp records with `now()` via `pack`; records encoded here will silently carry `t=0` (Unix epoch 1970).

pkg/senml/senml.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Decode parses a SenML JSON payload into a slice of records.
+func Decode(data []byte) ([]senml.Record, error) {
+	p, err := senml.Decode(data, senml.JSON)
+	if err != nil {
+		return nil, fmt.Errorf("senml decode: %w", err)
+	}
+	return p.Records, nil
```
JeffMboya, line 61: `senml.Decode` does not normalize — base fields (`bn`, `bt`, `bu`) remain on the first record and are not resolved into subsequent records. For a multi-record payload, `records[n].Name` (n>0) will be an empty string if the name was carried via `BaseName` on the first record. Call `senml.Normalize(p)` before returning `p.Records`.

rodneyosodo (approved):

PR #18, opened 2026-05-13
NOISSUE - Update heartbeat authentication

nyagamunene:
### What does this do?

Refactors the agent's heartbeat model to use direct MQTT instead of the internal PubSub broker:

- Removes the `messaging.PubSub` dependency from `agent.New` and the `broker` struct
- Agent publishes its own SenML heartbeat directly to `m/<domain>/c/<ctrl_channel>/services/agent/heartbeat` at startup and on a ticker
- Downstream services (e.g. Node-RED) do the same via the entrypoint heartbeat process
- `pkg/conn` handles incoming service heartbeats by calling `UpdateLiveness` on the agent service instead of forwarding to the internal broker
- Adds `UpdateLiveness(svcname, svctype string) error` to the `Service` interface
- Initial heartbeat is published immediately on startup (before the first ticker fires) so the agent appears in the services list without waiting a full interval

N/A

- **`agent.New` signature change**: the `messaging.PubSub` parameter has been removed. Callers must update.
- **`conn.NewBroker` signature change**: the `messaging.PubSub` parameter has been removed. Callers must update.
- **`agent.Service` interface gains `UpdateLiveness(svcname, svctype string) error`**: any external implementations must add this method.
- **Config-save notification removed**: the old `saveConfig` published to `commands.export.config` via the internal broker after writing the file. That notification no longer fires. Any downstream system that subscribed to that topic to react to config changes will silently stop receiving it.

Yes — existing service and conn tests cover the updated paths.

Heartbeat topic format is `m/<domain_id>/c/<ctrl_channel_id>/services/<svcname>/heartbeat` with a SenML payload containing a `service_type` field.

### Notes

The config-save notification removal is intentional: the internal broker was the only mechanism for that publish and has been removed. If re-notification is needed in future it should go over MQTT directly.


scripts/entrypoint.sh
```diff
@@ -281,4 +285,59 @@ fs.writeFileSync(flowFile, `${JSON.stringify(nodes, null, 4)}\n`);
...
+const domainID    = process.env.MG_AGENT_DOMAIN_ID    || "";
+const ctrlChannel = process.env.MG_AGENT_CTRL_CHANNEL || "";
+
+if (!host || !domainID || !ctrlChannel) {
+  process.stderr.write("Incomplete MQTT config, heartbeat disabled\n");
+  process.exit(0);
+}
+
+// Publish at 80% of the agent heartbeat interval to stay safely within the window.
+const rawInterval = (process.env.MG_AGENT_HEARTBEAT_INTERVAL || "10s").replace(/[^0-9]/g, "");
```
rodneyosodo: **Node-RED heartbeat interval parsing is incorrect for non-second durations.**

The regex `replace(/[^0-9]/g, "")` strips all non-numeric characters, which breaks any duration that isn't in seconds:

- `"1m"` → `1` second (should be 60s)
- `"2h"` → `2` seconds (should be 7200s)
- `"1m30s"` → `130` seconds (should be 90s)

This means if `MG_AGENT_HEARTBEAT_INTERVAL` is set to a minute-or-hour-based duration, the Node-RED heartbeat will fire far too frequently (potentially every 1-2 seconds instead of 60-7200 seconds). The `publishMs` floor of 1000ms won't protect against this since the parsed value will still be >= 1.

Consider parsing the duration properly in JS (e.g. handle `m`/`h` suffixes), or at minimum document that the interval must be specified in seconds (e.g. `"30s"`).

service.go
```diff
@@ -694,6 +639,45 @@ func (a *agent) Publish(t, payload string) error {
...
+	}
+
+	publish := func() {
+		token := a.mqttClient.Publish(topic, qos, false, payload)
+		token.Wait()
+		if err := token.Error(); err != nil {
+			a.logger.Warn("self-heartbeat publish failed", slog.Any("error", err))
+		}
+	}
+	ticker := time.NewTicker(interval)
```
rodneyosodo, line 661: **No initial heartbeat publish on startup.**

The ticker only fires after the first `interval` elapses, so the agent won't appear in its own services list (or any subscriber's list) for the full duration after startup. For long intervals (e.g. 1h), this is a significant delay.

Consider publishing once immediately before entering the select loop:

```go
publish()
ticker := time.NewTicker(interval)
```

service.go
```diff
@@ -639,7 +584,7 @@ func (a *agent) saveConfig(ctx context.Context, service, fileName, fileCont stri
 		return errNoSuchService
```
rodneyosodo, line 584: **Behavioral change: config save no longer publishes a notification.**

The old code published to `commands.{service}.config` after saving. This was removed along with the `PubSub` dependency, but any downstream system that relied on this publish to react to config changes will silently stop receiving those notifications.

If this is intentional, it should be called out in the PR description under "changes that modify/break current functionality."

pkg/conn/conn.go
```diff
@@ -99,29 +93,41 @@ func (b *broker) Resubscribe() {
...
+	if len(channelParts) < 4 || channelParts[3] == "" {
+		return "", "", false
 	}
-	channelParts := channelPartRegExp.FindStringSubmatch(topic)
-	if len(channelParts) < 4 {
-		return ""
+	parts := filter.Drop(strings.Split(channelParts[3], "/"), isEmpty).([]string)
+	if len(parts) < 2 || parts[len(parts)-1] != "heartbeat" {
+		return "", "", false
 	}
```
rodneyosodo, line 114: **`parts[0]` is fragile for extracting the service name from the topic path.**

Currently, the topic format is `services/{svcname}/heartbeat`, so `parts[0]` works. But if the topic convention ever includes sub-paths (e.g. `services/namespace/nodered/heartbeat`), this would incorrectly extract `"namespace"` instead of `"nodered"`.

Using `parts[len(parts)-2]` would be more robust since it always takes the segment just before `"heartbeat"`, which the preceding check already guarantees is at `parts[len(parts)-1]`.

rodneyosodo (approved):

PR #22, opened 2026-05-14
NOISSUE - OTA binary updater

nyagamunene:
### What does this do?

Adds an OTA (over-the-air) binary updater to the Linux gateway agent.

- New `pkg/ota` package with a 6-state machine: `IDLE → TRIGGERED → DOWNLOADING → VERIFYING → READY → RESTARTING`
- HTTPS download with progress reported to `m/<domain>/c/<ctrl_ch>/ota/status` every 5% in the format `downloading: 42%`
- Optional SHA-256 verification: inline via trigger payload (`sha256:<hex>`) or fallback to a `.sha256` sidecar file
- Atomic binary replacement via `os.Rename` with cross-device fallback (copy + rename)
- In-place process restart via `syscall.Exec` — same PID, no service manager required
- `ParseTrigger` parses the comma-delimited trigger payload: `url[,sha256:<hex>][,<size>]`
- OTA triggered via MQTT command on the control channel:
  ```
  m/<domain>/c/<ctrl_ch>/req
  [{"bn":"agent:", "n":"ota", "vs":"https://releases.example.com/agent-v2.bin,sha256:abcdef...,153600"}]
  ```
- Config: `MG_AGENT_OTA_ENABLED`, `MG_AGENT_OTA_BINARY_PATH`, `MG_AGENT_OTA_DOWNLOAD_DIR`

N/A

- `agent.Service` interface gains a new `OTA(ctx context.Context, url string) error` method — any external implementations of the interface must add this method
- `agent.NewConfig` signature gains an `OTAConfig` parameter — callers must pass `agent.OTAConfig{}` as the last argument

Yes. `pkg/ota/ota_test.go` covers 25 test cases across:
- State machine string representations
- `ParseTrigger` — 9 cases including field ordering, whitespace tolerance, and error inputs
- Download — success, server errors, context cancellation, 5% progress granularity
- Verify — inline hash match/mismatch, sidecar match/mismatch/absent, inline takes precedence over sidecar
- Replace — atomic rename, executable bit, missing source error
- Run error paths — connection refused, server error, hash mismatch with temp file cleanup, state progression

Configuration variables documented in `docker/.env`. The `syscall.Exec` success path (process replacement) is intentionally not covered by unit tests — it would replace the test runner process.

### Notes

MQTT stream mode (chunked binary over `m/<d>/c/<ctrl_ch>/ota`) is deferred — Aeolus has the same gap. The trigger payload format (`url,sha256:hash,size`) is aligned with the Aeolus OTA spec.


rodneyosodo (commented):
## Review Summary

Overall this is a well-structured OTA update implementation. The state machine, atomic replace with cross-device fallback, and test coverage are solid. I have a few actionable items below.

---

### Bugs

**1. No URL scheme/protocol validation on the download URL**

\`pkg/ota/ota.go\` \`Run()\` passes the URL directly to \`http.NewRequestWithContext\`. An attacker with MQTT publish access could supply a \`file://\` URL (or another supported scheme) to read arbitrary local files and write them to the binary path, achieving remote code execution without any network fetch. This is mitigated by MQTT auth, but the blast radius is high.

Consider validating that the URL scheme is \`http\` or \`https\` before downloading, e.g.:

```go
if u, err := url.Parse(rawURL); err != nil || (u.Scheme != "http" && u.Scheme != "https") {
    return fmt.Errorf("ota: unsupported URL scheme %q", u.Scheme)
}
```

**2. \`Trigger.Size\` is parsed but never validated or used**

\`TriggerFromRecords\` parses the \`size\` field into \`Trigger.Size\`, but \`Run()\` never checks it against the downloaded file size. If the intent is to enforce a size limit (e.g., prevent disk exhaustion from a malicious server), the check is missing. If the field is not intended to be enforced yet, it should be documented as such.

---

### Potential Issues

**3. OTA runs synchronously on the MQTT handler goroutine**

In \`pkg/conn/conn.go:197-206\`, the OTA command calls \`b.svc.OTA(b.ctx, trigger.URL, trigger.SHA256Hex)\` directly in the MQTT message handler. Since OTA involves a potentially long download (up to 10 minutes), this blocks the MQTT message handler for the entire duration, preventing processing of any other commands on that subscription. Consider running OTA in a separate goroutine.

**4. Concurrent OTA triggers are not guarded**

There is no mutex or state check to prevent a second OTA trigger from starting while one is already in progress. Two concurrent OTA runs could race on the same binary path and download directory, leading to corruption.

**5. \`verify\` silently skips when both inline hash and sidecar are absent**

\`pkg/ota/ota.go\` \`verify()\` returns \`nil\` when no hash is available, meaning a downloaded binary with no integrity check is accepted and installed. For an OTA system that replaces the running binary, this is a significant security gap. Consider at minimum logging a warning, or requiring a hash when OTA is enabled.

**6. \`TriggerFromRecords\` does not validate the URL is non-empty early enough**

The empty-string check \`*r.StringValue == ""\` only fires if \`r.Name == "url"\`. But the final \`if t.URL == ""\` catch-all allows any non-empty string through, including whitespace-only values like \`" "\`. A whitespace URL will cause a confusing error downstream from \`http.NewRequestWithContext\`.

---

### Minor

**7. \`ota.Config.Enabled\` is redundant**

\`ota.Config.Enabled\` is copied from \`agent.OTAConfig.Enabled\` but is never checked inside \`ota.Run()\` — the enabled check happens in \`service.go\` before calling \`Run()\`. The field in \`ota.Config\` serves no purpose and could be removed for clarity.

**8. Logging middleware logs \`sha256hex\` in the URL field**

\`middleware/logging.go\` only logs the \`url\` parameter but not \`sha256hex\`, making it harder to correlate logs with specific OTA requests. Consider logging both parameters.

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	if !verified {
+		slog.Default().Warn("OTA binary installed without integrity verification; no hash provided and no sidecar found")
+	}
+
+	progressFn(StateReady, 100)
+	if err := replace(tmpPath, cfg.BinaryPath); err != nil {
+		_ = os.Remove(tmpPath)
+		return fmt.Errorf("ota replace: %w", err)
+	}
```
rodneyosodo, line 135: **Bug: No URL scheme validation.** Any scheme is accepted (e.g. `file:///etc/passwd`, `ftp://`, `gopher://`). An attacker with MQTT access could use `file://` to read arbitrary local files into `DownloadDir`, or write crafted content. Should validate that `url` has scheme `http://` or `https://` before calling `http.NewRequestWithContext`.

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+			if r.Value != nil {
+				t.Size = uint64(*r.Value)
+			}
+		}
+	}
+	if t.URL == "" {
+		return Trigger{}, fmt.Errorf("ota trigger: url is required")
+	}
+	return t, nil
+}
```
rodneyosodo, line 103: `Trigger.Size` is parsed here but never used anywhere — not in `download()` to pre-allocate or verify, not in `verify()`, not in `Run()`. This is dead code. Either enforce the expected byte count (e.g. error if `written > t.Size`) or remove the `Size` field and its parsing to avoid confusion.

pkg/conn/conn.go
```diff
@@ -187,5 +194,22 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 		if err := syscall.Exec(os.Args[0], os.Args, os.Environ()); err != nil {
 			b.logger.Error("Reset failed", slog.Any("error", err))
 		}
+	case otaCmd:
+		trigger, err := ota.TriggerFromRecords(sm.Records[1:])
+		if err != nil {
```
rodneyosodo, line 197: **OTA blocks the MQTT handler goroutine.** `b.svc.OTA()` runs synchronously inside `handleMsg()`, which is the MQTT message callback. The HTTP client has a 10-minute timeout, so the agent will be unable to process any other MQTT messages (heartbeats, commands, pings) during the entire download+verify+replace cycle.

Consider running this in a goroutine:
```go
go func() {
    if err := b.svc.OTA(context.Background(), trigger.URL, trigger.SHA256Hex); err != nil {
        b.logger.Warn("OTA operation failed", slog.Any("error", err))
    }
}()
```

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+	lastPct := 0.0
+	buf := make([]byte, 32*1024)
+	pctFn(0)
+
+	for {
+		if ctx.Err() != nil {
+			f.Close()
+			os.Remove(tmpName)
+			return "", ctx.Err()
+		}
```
rodneyosodo, line 186: **Verification silently skipped when no hash is provided.** When `sha256hex` is empty and the sidecar fetch fails (network error, non-200 status), the function returns `nil` — meaning the binary is installed with zero integrity verification and no log output. This is a security concern.

Consider either:
1. Requiring at least one hash source and returning an error if neither is available.
2. At minimum, logging a warning that verification was skipped.

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+//
+// Pass sm.Records[1:] to skip the dispatch record.
+func TriggerFromRecords(records []senml.Record) (Trigger, error) {
+	var t Trigger
+	for _, r := range records {
+		switch r.Name {
+		case "url":
+			if r.StringValue == nil || strings.TrimSpace(*r.StringValue) == "" {
+				return Trigger{}, fmt.Errorf("ota trigger: url record has no value")
+			}
```
rodneyosodo, line 87: Whitespace-only URL values (e.g. `"  "`) pass this check since they are not `""`. Should `strings.TrimSpace` the value before checking, e.g.:
```go
t.URL = strings.TrimSpace(*r.StringValue)
```
and then check against `""`.

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+
+	progressFn(StateTriggered, 0)
+
+	progressFn(StateDownloading, 0)
+	tmpPath, err := download(ctx, url, cfg.DownloadDir, func(pct float64) {
+		progressFn(StateDownloading, pct)
+	})
+	if err != nil {
+		return fmt.Errorf("ota download: %w", err)
+	}
```
rodneyosodo, line 119: **No concurrency guard.** If two OTA triggers arrive in quick succession, two concurrent `Run()` calls could both download and attempt `replace()` on the same binary path, leading to a race condition. Consider adding a `sync.Mutex` or atomic flag to serialize OTA runs.

pkg/ota/ota.go
```diff
@@ -0,0 +1,319 @@
...
+// sha256hex is the expected SHA-256 hex digest; if empty the sidecar at url+".sha256" is tried.
+// size is the expected byte count; if non-zero the downloaded file is checked against it.
+// On success it never returns (the process is replaced via syscall.Exec).
+// On any failure it returns a descriptive error; the running binary is untouched.
+func Run(ctx context.Context, cfg Config, url, sha256hex string, size uint64, progressFn ProgressFn) error {
+	u, err := neturl.Parse(url)
+	if err != nil {
+		return fmt.Errorf("ota: invalid URL: %w", err)
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
```
rodneyosodo: `Enabled` field is never read inside the `ota` package. The enabled check is done in `service.go:OTA()` before constructing this config, so this field is redundant. Consider removing it to avoid the implication that the `ota` package checks it.

middleware/logging.go
```diff
@@ -180,6 +180,24 @@ func (lm *loggingMiddleware) UpdateLiveness(svcname, svctype string) (err error)
 	return lm.svc.UpdateLiveness(svcname, svctype)
 }
 
+func (lm *loggingMiddleware) OTA(ctx context.Context, url, sha256hex string, size uint64) (err error) {
+	defer func(begin time.Time) {
+		args := []any{
+			slog.String("duration", time.Since(begin).String()),
```
rodneyosodo, line 186: `sha256hex` is accepted as a parameter but not included in the log args. This makes it harder to correlate log lines with verification failures. Consider adding:
```go
slog.String("sha256hex", sha256hex)
```

rodneyosodo (commented):
## Review Summary

Reviewing the OTA binary update feature introduced in this PR.

### Bugs
1. **No URL scheme validation** — `file://`, `ftp://`, or other schemes could be used to read/write arbitrary files. Should validate that the URL scheme is `http://` or `https://` before proceeding.
2. **`Trigger.Size` is parsed but never used** — dead code; either enforce the size limit or remove the field.

### Potential Issues
3. **OTA blocks the MQTT handler goroutine** — `b.svc.OTA()` runs synchronously inside `handleMsg()`, blocking message processing for up to 10 minutes (the HTTP client timeout). Consider running in a separate goroutine.
4. **No concurrency guard** — Two rapid OTA triggers could race, both downloading and attempting to replace the same binary.
5. **Verification silently skipped** — When no hash is provided and the sidecar is unreachable, the binary is installed without integrity verification. Consider making verification mandatory or at least logging a warning.
6. **Whitespace-only URL values** — A URL of `"  "` passes the `StringValue != nil && *r.StringValue == ""` check but is not a valid URL.

### Minor
7. **`ota.Config.Enabled` is redundant** — The enabled check is done in `service.go:OTA()` before constructing the config, so `ota.Config.Enabled` is never read inside the `ota` package.
8. **Logging middleware omits `sha256hex`** — The logging middleware logs the URL but not the hash, making it harder to debug verification failures.

rodneyosodo (approved):

PR #23, opened 2026-05-18
NOISSUE - Add Device manager

nyagamunene:
### What does this do?

Adds a downstream device registry and Magistrala Provision API client to the Linux gateway agent.

- New `pkg/devicemgr` package:
  - `Device` struct: `id`, `key`, `channel_id`, `interface_type` (ble/serial/i2c/usb/zigbee), `interface_addr`, `name`, `active`, `last_seen`
  - BoltDB-backed persistent store (`store.go`) — survives restarts; operations: `Save`, `Get`, `Remove`, `List`, `MarkSeen`
  - Magistrala Provision API client (`provision.go`) — `POST /<domain>/mapping` with Bearer token or Client key auth
  - `Manager` (`manager.go`) composing store + provision: `Add`, `Remove`, `Get`, `List`, `MarkSeen`
- `DeviceManager(uuid, cmdStr string) error` added to the `agent.Service` interface, wired to the `devices` MQTT command
- Supported subcommands:
  - `list` — returns a JSON array of all registered devices
  - `add,<name>,<ext_id>,<ext_key>,<iface_type>,<iface_addr>` — provision via Magistrala and persist locally
  - `remove,<device_id>` — deregister a device
  - `get,<device_id>` — fetch a single device record
  - `seen,<device_id>` — update `last_seen` and mark `active = true`
- New env vars: `MG_AGENT_PROVISION_URL`, `MG_AGENT_PROVISION_TOKEN`, `MG_AGENT_DEVICE_DB_PATH` (default: `/var/lib/agent/devices.db`)
- Device manager is optional — omitting `MG_AGENT_PROVISION_URL` disables it gracefully
- Docker: `agent-data` volume added to persist the BoltDB file across container restarts

N/A

- `agent.Service` interface gains `DeviceManager(uuid, cmdStr string) error` — any external implementations must add this method
- `agent.New` gains a `*devicemgr.Manager` parameter (pass `nil` to disable)

Yes. Tests added across three files:

`pkg/devicemgr/store_test.go`:
- `NewStore` valid and invalid path
- `Save`+`Get` round-trip and overwrite behaviour
- `Remove` existing and non-existent (no-op)
- `List` empty and populated store
- `MarkSeen` sets `Active = true`, updates `LastSeen`, errors on missing device

`pkg/devicemgr/manager_test.go`:
- `New` valid/invalid DB path
- `Add` success, empty provision URL, server error, empty clients response, unreachable server
- Auth header — Bearer token when configured, `Client <key>` fallback
- `Remove`, `Get`, `List`, `MarkSeen` happy and error paths
- `ParseInterfaceType` all known types + unknown/empty

`service_test.go`:
- Disabled manager returns error
- Full lifecycle: list empty → add → list → get → seen → remove → get-after-remove
- Error paths: empty command, unknown subcommand, missing arguments for each subcommand

Configuration variables documented in `docker/.env`. Relationship to `provision.sh` is intentional: the shell script provisions the **gateway itself** (one-time operator setup); `provision.go` provisions **downstream devices** managed by the gateway at runtime.

### Notes

Design aligned with the Aeolus (`device_manager.c`) spec: same `Device` fields, same lifecycle operations. The BoltDB store replaces Aeolus's NVS-packed colon-separated string persistence with JSON-encoded records.


dborovcanin:
@rodneyosodo Please review.

pkg/devicemgr/manager.go
```diff
@@ -0,0 +1,153 @@
...
+	"time"
+
+	"github.com/absmach/agent/pkg/iface"
+)
+
+// Manager handles the downstream device registry and Magistrala provisioning.
+type Manager struct {
+	store      *Store
+	provision  *provisionClient
+	interfaces map[string]iface.Interface
```
rodneyosodo, line 22: `interfaces` is mutated from `OpenIface`, `CloseIface`, `Remove`, and `Close`, and read from `ReadIface`/`WriteIface`. Once the MQTT goroutine is running `DeviceManager → ReadIface` and another command issues `remove`, you have concurrent map access — the Go runtime will throw `concurrent map writes` (not a recoverable error). Wrap with a `sync.Mutex` and snapshot the iface pointer under the lock before doing the actual Read/Write/Close I/O outside it.

pkg/devicemgr/manager.go
```diff
@@ -0,0 +1,153 @@
...
+		ifaceCfg:   ifaceCfg,
+	}, nil
+}
+
+// Close releases all open interfaces and the store.
+func (m *Manager) Close() error {
+	for id, ifc := range m.interfaces {
+		_ = ifc.Close()
+		delete(m.interfaces, id)
+	}
```
rodneyosodo, line 53: Errors from `ifc.Close()` are silently dropped on shutdown. At minimum log them — a stuck interface during shutdown is exactly the kind of thing you want visible in logs. Or collect with `errors.Join` and return.

pkg/devicemgr/manager.go
```diff
@@ -0,0 +1,153 @@
...
+	return m.store.Close()
+}
+
+// Add provisions a new downstream device via the Magistrala Provision API,
+// saves it to the local registry, and returns it.
+func (m *Manager) Add(name, externalID, externalKey string, ifaceType iface.InterfaceType, ifaceAddr string) (Device, error) {
+	if m.provision.cfg.URL == "" {
```
rodneyosodo: The `cfg.URL == ""` check duplicates the guard documented in `New()`. Move the check into `provisionClient.Provision()` itself so the contract is enforced in one place (and so a future `Add` path doesn't accidentally skip it).

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,115 @@
...
+
+type provisionClient struct {
+	cfg    ProvisionConfig
+	client *http.Client
+}
+
+func newProvisionClient(cfg ProvisionConfig) *provisionClient {
+	return &provisionClient{
+		cfg:    cfg,
+		client: &http.Client{Timeout: 20 * time.Second},
```
rodneyosodo, line 34: 20 s hard timeout with no `context.Context`. Provision is invoked from the MQTT command goroutine — if the agent is shutting down or the operator wants to cancel, there's no mechanism. Take `ctx` as the first argument of `Provision` (and `Add`/`DeviceManager` up the stack), and use `http.NewRequestWithContext`.

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,115 @@
...
+
+	req, err := http.NewRequest(http.MethodPost, endpoint, bytes.NewReader(body))
+	if err != nil {
+		return Device{}, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if p.cfg.Token != "" {
+		req.Header.Set("Authorization", "Bearer "+p.cfg.Token)
+	} else {
+		req.Header.Set("Authorization", "Client "+externalKey)
```
rodneyosodo: Fallback `Authorization: Client <externalKey>` is using the *device's own pre-shared key* as the credential for provisioning that same device. Most Magistrala Provision deployments don't accept this auth scheme, and even where they do, it muddies the security model — anyone who knows a device's key can register additional clients/channels under your domain. Either require a Token and return an explicit error otherwise, or document this contract precisely.

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	resp, err := p.client.Do(req)
+	if err != nil {
+		return Device{}, fmt.Errorf("provision request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
+		b, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
```
rodneyosodo: Good that error bodies are size-limited via `io.LimitReader`, but the success-path body on line 96 is unbounded. If the server (or anyone in the middle) returns a giant body, you'll happily decode the whole thing. Use `io.LimitReader(resp.Body, …)` on the success path too.

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,115 @@
...
+		return Device{}, fmt.Errorf("provision API returned %d: %s", resp.StatusCode, b)
+	}
+
+	var pr provisionResponse
+	if err := json.NewDecoder(resp.Body).Decode(&pr); err != nil {
+		return Device{}, fmt.Errorf("decode provision response: %w", err)
+	}
+	if len(pr.Clients) == 0 {
```
rodneyosodo: Only `Clients[0]` and `Channels[0]` are read. If the Provision API ever returns multiple entries (bulk path, or a future multi-channel response), the rest are silently dropped. Either assert `len == 1` and error otherwise, or document the contract.

pkg/devicemgr/store.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	return s.db.Update(func(tx *bolt.Tx) error {
+		return tx.Bucket(devicesBucket).Put([]byte(d.ID), b)
+	})
+}
+
+// Remove deletes a device by its ID. No error if the device is absent.
+func (s *Store) Remove(id string) error {
+	return s.db.Update(func(tx *bolt.Tx) error {
+		return tx.Bucket(devicesBucket).Delete([]byte(id))
```
rodneyosodo, line 56: BoltDB `Delete` returns nil for a missing key, so the `Remove` of a non-existent device is silently a no-op. The PR description calls this out as intentional — worth mirroring that in the godoc here (`No error if the device is absent.` is on the right line — good).

pkg/iface/iface.go
```diff
@@ -0,0 +1,88 @@
...
+		c := cfg.Modbus
+		c.Mode = modbus.ModeTCP
+		c.Address = addr
+		return modbus.New(c), nil
+	case InterfaceI2C:
+		c := cfg.I2C
+		c.Bus = addr
+		return i2c.New(c), nil
+	default:
+		return nil, fmt.Errorf("unsupported interface type: %s", ifaceType)
```
rodneyosodo, line 90: `InterfaceBLE` and `InterfaceZigbee` are first-class enum values (parsable, mentioned in the godoc above) but `New()` rejects them with `"unsupported interface type"`. End-to-end, a user provisions a BLE device successfully and only fails when something tries to open the iface. Either implement a stub that returns a clear `not implemented` from `Open()`, or remove these values from the enum until they're real.

pkg/iface/i2c/i2c.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	n, err := d.f.Write(data)
+	if err != nil {
+		return n, fmt.Errorf("i2c: write: %w", err)
+	}
+	return n, nil
+}
+
+// ensure unsafe is used (required by the ioctl call path via unix package).
+var _ = unsafe.Sizeof(0)
```
rodneyosodo: This trailing `var _ = unsafe.Sizeof(0)` is a workaround to silence the unused-import warning, but nothing in this file actually uses `unsafe` — `unix.IoctlSetInt` takes an `int`, not an unsafe pointer. Drop both the `unsafe` import and this line.

pkg/iface/i2c/i2c_unsupported.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// New returns an I2C stub that always errors.
+func New(cfg Config) *I2C {
+	return &I2C{cfg: cfg}
+}
+
+func (d *I2C) Open() error                { return fmt.Errorf("i2c: not supported on this platform") }
+func (d *I2C) Close() error               { return nil }
+func (d *I2C) Read(buf []byte) (int, error)  { return 0, fmt.Errorf("i2c: not supported on this platform") }
```
rodneyosodo: Inconsistent argument-list alignment in this block (`Open()`, `Close()`, `Read(buf []byte)`, `Write(data []byte)`). `gofmt` should normalise this — please run it.

service.go
```diff
@@ -750,3 +763,141 @@ func (a *agent) getTopic(topic string) (t string) {
...
+//	open,<device_id>                             → open physical interface
+//	close,<device_id>                            → close physical interface
+//	read,<device_id>,<n_bytes>                   → read n bytes, reply as hex
+//	write,<device_id>,<hex_data>                 → write hex bytes to interface
+func (a *agent) DeviceManager(uuid, cmdStr string) error {
+	if a.devices == nil {
+		return errDeviceManagerDisabled
+	}
+
+	args := strings.Split(strings.TrimSpace(cmdStr), ",")
```
rodneyosodo, line 800: Splitting `cmdStr` on `,` makes any subcommand argument that contains a comma — most plausibly a human-friendly device `name` — silently desync the arg positions. The result is a provisioning call with the wrong fields populated and no error. Either validate `name`/`ext_id`/etc. server-side, or switch to a JSON payload (the `Control`/`ServiceConfig` MQTT commands have the same fragility, but adding a new comma-delimited surface compounds it).

service.go
```diff
@@ -750,3 +763,141 @@ func (a *agent) getTopic(topic string) (t string) {
...
+			return errors.Wrap(errDeviceManagerFailed, err)
+		}
+		resp = "ok"
+
+	case "read":
+		if len(args) < 3 {
+			return errors.Wrap(errDeviceManagerFailed, errInvalidCommand)
+		}
+		var n int
+		if _, serr := fmt.Sscanf(args[2], "%d", &n); serr != nil || n <= 0 {
```
rodneyosodo: Use `strconv.Atoi` here — `fmt.Sscanf("%d", &n)` is overkill for a single integer and a bit slower. Also: `n` has no upper bound, so `read,<id>,9999999999` allocates a 10 GB buffer in `ReadIface`. Cap the size.

pkg/conn/conn.go
```diff
@@ -204,5 +205,10 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 				b.logger.Warn("OTA operation failed", slog.Any("error", err))
 			}
 		}()
+	case devices:
+		b.logger.Info("Devices command", slog.String("uuid", uuid), slog.String("command", cmdStr))
+		if err := b.svc.DeviceManager(uuid, cmdStr); err != nil {
+			b.logger.Warn("DeviceManager operation failed", slog.Any("error", err))
```
rodneyosodo, line 213: When `DeviceManager` fails, the error is only logged locally — the remote operator who sent the command never sees it. The existing `Control` path has the same shape so this is consistent, but for the operator-facing device commands (`add`, `remove`, etc.) silent failures will lead to a lot of "why didn't anything happen?" — consider publishing a failure response back to the control topic, even just `{"error": "…"}`.

cmd/main.go
```diff
@@ -125,7 +130,22 @@ func main() {
 	}
 	noderedClient := nodered.NewClient(cfg.NodeRed.URL, logger)
 
-	svc, err := agent.New(ctx, mqttClient, &cfg, noderedClient, logger)
+	var devices *devicemgr.Manager
+	if c.ProvisionURL != "" {
```
rodneyosodo: Device manager is only constructed when `c.ProvisionURL != ""`. That means there's no way to ship a gateway with a pre-provisioned device list (or to use the device-manager API surface without ever calling `Add`). If that's intentional, document it on `MG_AGENT_DEVICE_DB_PATH` in `docker/.env`; otherwise, create the manager unconditionally and just fail `Add` when `ProvisionURL` is empty.

docker/docker-compose.yml
```diff
@@ -70,6 +71,9 @@ services:
       MG_AGENT_OTA_ENABLED: ${MG_AGENT_OTA_ENABLED}
       MG_AGENT_OTA_BINARY_PATH: ${MG_AGENT_OTA_BINARY_PATH}
       MG_AGENT_OTA_DOWNLOAD_DIR: ${MG_AGENT_OTA_DOWNLOAD_DIR}
+      MG_AGENT_PROVISION_URL: ${MG_AGENT_PROVISION_URL}
+      MG_AGENT_PROVISION_TOKEN: ${MG_AGENT_PROVISION_TOKEN}
+      MG_AGENT_DEVICE_DB_PATH: ${MG_AGENT_DEVICE_DB_PATH}
     volumes:
       - /etc/ssl/certs/ca-certificates.crt:/etc/ssl/certs/ca-certificates.crt:ro
-
+      - agent-data:/var/lib/agent
```
rodneyosodo: `/var/lib/agent` needs write permission for whatever user the agent runs as. If anyone later switches the agent image to a non-root user (recommended for production), the BoltDB open will fail with a confusing permission error. Either document the required UID/GID, switch the default path to something under the agent user's home, or ensure the volume is `chown`ed at container start.

docker/.env
```diff
@@ -33,3 +33,8 @@ MG_AGENT_BOOTSTRAP_URL=http://bootstrap:9013/clients/bootstrap
 MG_AGENT_BOOTSTRAP_RETRIES=5
 MG_AGENT_BOOTSTRAP_RETRY_DELAY_SECONDS=10
 MG_AGENT_BOOTSTRAP_SKIP_TLS=false
+
+## Device Manager
+MG_AGENT_PROVISION_URL=
+MG_AGENT_PROVISION_TOKEN=
+MG_AGENT_DEVICE_DB_PATH=/var/lib/agent/devices.db
```
rodneyosodo, line 38: Treating `MG_AGENT_PROVISION_TOKEN` as a plain-text env var means it lands in `docker inspect`, container logs (if anyone prints env), and any host backup of `.env`. The token grants device-provisioning rights against your domain. At least mention this in the comment; longer-term, prefer a Docker secret or the bootstrap-delivered token approach in PR #24.

pkg/devicemgr/manager_test.go
```diff
@@ -0,0 +1,408 @@
...
+		{"", iface.InterfaceUnknown},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.input, func(t *testing.T) {
+			got := iface.ParseInterfaceType(tc.input)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
```
rodneyosodo: `TestParseInterfaceType` here duplicates the identical test in `pkg/iface/iface_test.go`. Keep one — probably the iface-package version, since `ParseInterfaceType` lives there.

rodneyosodo (commented):
## Overall

This PR introduces a clean `pkg/devicemgr` (BoltDB store + Magistrala Provision HTTP client + Manager) and a new `DeviceManager` service method wired to the existing `devices` MQTT command. The scope is well-contained, tests cover the store and manager solidly, and the optional-when-`MG_AGENT_PROVISION_URL`-is-empty pattern is good defensive design.

Main concerns:

**Concurrency**
- `pkg/devicemgr/Manager.interfaces` is a plain `map[string]iface.Interface` mutated from `Open/Close/Read/WriteIface`, `Remove`, and `Close`. The MQTT goroutine calls `DeviceManager → OpenIface/ReadIface` while other handlers can run `Remove` concurrently. Result: race detector will flag `concurrent map writes` immediately under any real load. Add a `sync.Mutex`; snapshot the iface pointer under the lock before doing I/O outside it.

**Provisioning client**
- `provision.go` falls back to `Authorization: Client <externalKey>` when no Token is configured. That sends the *device's own pre-shared key* as the provisioning credential — most Magistrala Provision deployments do not accept this scheme, and even if they do, it conflates device identity with operator authority. Either require `Token` and return an explicit error otherwise, or document the contract precisely.
- `provision.go` uses `http.Client{Timeout: 20s}` with no plumbed `context.Context`. A caller (HTTP handler, MQTT command, agent shutdown) cannot cancel an in-flight provision. Take `ctx` as the first argument and use `http.NewRequestWithContext`.
- The response decoder unconditionally reads `Clients[0]` / `Channels[0]`; if the Provision API ever returns multiple clients (e.g. bulk provisioning), data is silently dropped without a warning.

**API surface / consistency**
- `pkg/iface/iface.go` lists `InterfaceBLE` and `InterfaceZigbee` as first-class enum values, but `iface.New()` returns *"unsupported interface type"* for both. A user `add`s a BLE device, succeeds on provision (the Magistrala side doesn't care about the bus), and only discovers it doesn't work when something tries to open the interface. Either implement stubs that return a clear `not implemented`, or drop these values from the enum.
- `service.go` `DeviceManager` parses `cmdStr` by splitting on `,`. A device `name` containing a comma corrupts the arg positions. Validate `name`/`ext_id`/etc. server-side, or use a JSON payload.
- `pkg/iface/i2c/i2c.go` ends with `var _ = unsafe.Sizeof(0)` — a no-op that exists only to silence the unused-import warning, but nothing in the file actually needs `unsafe`. Drop the import and the line.

**Operational / packaging**
- `cmd/main.go` enables the device manager only when `c.ProvisionURL != ""`. There's no way to ship a gateway with a pre-provisioned device list (or a future no-provisioning path) — `MG_AGENT_DEVICE_DB_PATH` is silently ignored if no URL is configured.
- `docker/docker-compose.yml` mounts the named volume to `/var/lib/agent` — that path needs write permission for the agent process. If anyone later runs the container as a non-root user (good practice), the BoltDB open will fail. Worth a comment or a tighter default path in the agent image.

**Tests**
- `pkg/devicemgr/manager_test.go::TestParseInterfaceType` and `pkg/iface/iface_test.go::TestParseInterfaceType` are line-for-line duplicates. Keep one (probably in the `iface` package, where the function lives).
- The new tests are clean and table-driven — nice. Consider one race-tagged test that runs `OpenIface`/`Remove` concurrently to catch the map race noted above.

Detailed inline comments below.

rodneyosodo (approved):

PR #24, opened 2026-05-21
NOISSUE - Add OTA and device manager to UI

nyagamunene:
### What does this do?

Adds a downstream device manager, SDK-based provisioning, a telemetry scheduler, and Devices/OTA UI pages to the Linux gateway agent.

- **Device Manager** (`pkg/devicemgr`): BoltDB-backed registry with Add, Remove, Get, List, MarkSeen, FindByAddr, and interface open/read/write. `FindByAddr` mirrors the Aeolus (Zephyr) companion agent for cross-platform consistency.
- **SDK provisioning** (`provision.go`): Replaces the Provision service `/mapping` call with direct SDK steps — CreateClient → CreateChannel → Connect → AddRule (`save_senml`) — with rollback on failure at each step.
- **Bootstrap integration**: Profile template now delivers `clients_url`, `channels_url`, `rules_engine_url`, and `token` to the agent at startup so no device-manager env vars are needed in docker-compose.
- **Telemetry Scheduler** (`scheduler.go`): Goroutine-per-device that connects to MQTT using each device's own credentials (required by Magistrala ACLs) and publishes to `m/<domainID>/c/<channelID>/msg`, with 5 s retry backoff.
- **provisioning script**: `provision.sh` separates script-facing URLs (`localhost`) from agent-facing URLs (`clients:9006`, `channels:9005`, `rules:9008`) so the script runs on the host while the agent resolves services via Docker `extra_hosts`.
- **UI**: New Devices page (register downstream devices, copy channel ID) and OTA page; removed `"(no output)"` from the Execute terminal.

N/A

- `ProvisionConfig` fields renamed: `URL` → `ClientsURL`, added `ChannelsURL`, `RulesURL` → `RulesEngineURL`. Switch env var from `MG_AGENT_PROVISION_URL` to `MG_AGENT_CLIENTS_URL` + `MG_AGENT_CHANNELS_URL`. Bootstrap-based deployments are unaffected.
- Bootstrap profile template format changed (`provision.url` → `provision.clients_url` / `channels_url` / `rules_engine_url`). Re-run `provision.sh` to create a new enrollment.

Yes — full unit tests for the store, manager, and `FindByAddr` in `pkg/devicemgr`, with manager tests mocking the Magistrala SDK HTTP endpoints directly. `service_test.go` covers device manager service methods end-to-end.

`provision.sh` summary output now shows script-facing and agent-facing URLs separately.

### Notes

- Physical interface reads in the scheduler are currently stubs; full BLE/Serial read loops will follow in a subsequent PR.


rodneyosodo (commented):
## Overall

This PR adds a meaningful new surface (downstream device registry, SDK-based provisioning, per-device telemetry scheduler, Devices + OTA UI). The structure is sensible and well-tested at the unit level — `pkg/devicemgr` tests against an `httptest` server are a nice touch.

The main concerns are concurrency safety and a few HTTP-level bugs:

**Blocking / correctness**
- `api/endpoints.go` `otaTriggerEndpoint` launches a goroutine bound to the **request context**, which is cancelled the moment the HTTP response is written. The OTA will be aborted before it does any meaningful work. Use `context.Background()` (or a long-lived `a.ctx`).
- `pkg/devicemgr/Manager.interfaces` map is mutated from HTTP handlers (`AddDevice` → no, but `OpenIface`/`Remove`) **and** from the scheduler goroutine (`OpenIface`, `CloseIface`, `ReadIface`, `WriteIface`) with no lock. Concurrent map access — race detector will flag this immediately on a real device with parallel API + scheduler.
- `scheduler.go` stores `s.ctx` in `Start()` and reads it from `StartDevice()` without synchronisation; if `StartDevice` is called before `Start`, `s.ctx` is nil → `context.WithCancel(nil)` panics. Also a plain data race.
- `api/common.go` default branch in `EncodeError` calls `w.WriteHeader(500)` twice when the JSON encode fails — the second call is a no-op with a runtime warning. Same issue is *intentionally* avoided in the other branches (because the header is set before the encode), but here it's redundant and noisy.

**Design / consistency**
- New routes live under `/api/...` while every existing route (`/exec`, `/nodered`, `/services`, `/health`, …) is at the root. Pick one and stick with it.
- `pkg/iface/iface.go` lists `InterfaceBLE` and `InterfaceZigbee` in the enum but `New()` returns *"unsupported interface type"* for both. Either drop them from the type or stub them — currently a user picks BLE in the UI and gets a confusing error post-provision.
- `Service` interface now has six device methods + OTAStatus. Worth a sub-interface (`DeviceService`) embedded into `Service`.
- The DeviceManager MQTT command parser splits on `,` — a device name containing a comma silently corrupts the command. Either validate `name` server-side or use a JSON payload (as `Control`/`ServiceConfig` arguably should too).

**Security**
- `scripts/provision.sh` embeds `MG_PAT` in the bootstrap `render_context`. The PAT then traverses bootstrap → agent, granting the agent the user's full PAT-level authority. That's the simplest path but worth documenting clearly: the agent inherits the operator's privileges, and the PAT lives in the device store / on disk if the agent persists state. Consider scoping a per-device PAT.
- `pkg/devicemgr/provision.go` always uses a fresh `context.Background()` with a hard 30 s timeout, ignoring caller cancellation (cmd shutdown, request cancel). Inherit from a passed context.

**UI**
- `ui/src/pages/devices.tsx` `handleRemove` deletes without confirmation — easy mis-click.
- `ui/src/pages/ota.tsx` polls `/api/ota/status` every 3 s forever; consider stopping when `busy=false && !submitted` to reduce noise (and CPU/network on a tab left open).
- `ui/src/pages/ota.tsx` reads the error body as `res.text()` but the server returns JSON `{"error":"..."}`; the user sees the JSON string verbatim.

**Test coverage**
- Solid unit tests for the new store / manager. Scheduler has no tests — a fake `mqtt.Client` + `iface.Interface` would let you cover the retry loop and the `StopDevice` path.
- `pkg/devicemgr/manager_test.go` and `pkg/iface/iface_test.go` both have a `TestParseInterfaceType` — they are identical; pick one.

Detailed inline comments below.

api/common.go
```diff
@@ -87,5 +87,8 @@ func EncodeError(_ context.Context, err error, w http.ResponseWriter) {
 		return
 	default:
 		w.WriteHeader(http.StatusInternalServerError)
+		if err := json.NewEncoder(w).Encode(map[string]string{"error": err.Error()}); err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
```
rodneyosodo: `w.WriteHeader(http.StatusInternalServerError)` is already called on line 89 before the encode attempt. By the time the encoder fails, the status line and headers have been flushed, and this second `WriteHeader` is a no-op that logs `http: superfluous response.WriteHeader call`. Either drop the second call (the body is already broken; just log the error) or — to mirror the other branches — set the status only inside the `if`, after the encode fails. The other case branches in this file get it right; only `default` is double-writing.

api/endpoints.go
```diff
@@ -128,3 +129,81 @@ func nodeRedEndpoint(svc agent.Service) endpoint.Endpoint {
...
+}
+
+func otaTriggerEndpoint(svc agent.Service) endpoint.Endpoint {
+	return func(ctx context.Context, request any) (any, error) {
+		req := request.(otaTriggerReq)
+		if err := req.validate(); err != nil {
+			return nil, err
+		}
+		go func() {
+			_ = svc.OTA(ctx, req.URL, req.SHA256Hex, req.Size)
```
rodneyosodo: `ctx` here is the **request context**, which `net/http` cancels as soon as the response is written. The goroutine you're launching will see `ctx.Done()` almost immediately and the OTA will abort (depending on how strictly `ota.Run` honours ctx) before doing any real work — but you'll have already returned `"triggered"` to the client. Use `context.Background()` for the goroutine, or thread the agent's long-lived ctx into the service. The same shape applies to any future async endpoint.

api/transport.go
```diff
@@ -81,6 +81,51 @@ func MakeHandler(svc agent.Service, logger *slog.Logger, stream *logstream.Strea
 		opts...,
 	).ServeHTTP)
 
+	r.Route("/api", func(r chi.Router) {
```
rodneyosodo: Why `/api/...` for the new routes when every existing route (`/exec`, `/nodered`, `/services`, `/metrics`, `/health`, …) is at the root? Pick one — either move all newly added routes to root for consistency, or migrate the others under `/api` and call it v1. Mixing the two is a small but real cliff for anyone integrating against this surface.

api/requests.go
```diff
@@ -96,3 +96,31 @@ func (req nodeRedReq) validate() error {
...
+type addDeviceReq struct {
+	Name      string `json:"name"`
+	ExtID     string `json:"ext_id"`
+	ExtKey    string `json:"ext_key"`
+	IfaceType string `json:"interface_type"`
+	IfaceAddr string `json:"interface_addr"`
+}
+
+func (req addDeviceReq) validate() error {
+	if req.Name == "" || req.ExtID == "" || req.ExtKey == "" || req.IfaceType == "" {
```
rodneyosodo: `IfaceAddr` isn't required, but most interface types are useless without it (BLE needs a MAC, serial needs a path, modbus-tcp needs host:port). Either make it required, or validate per `IfaceType` here so the error surfaces at the HTTP boundary rather than deep inside `OpenIface` later.

pkg/devicemgr/manager.go
```diff
@@ -0,0 +1,157 @@
...
+	"time"
+
+	"github.com/absmach/agent/pkg/iface"
+)
+
+// Manager handles the downstream device registry and Magistrala provisioning.
+type Manager struct {
+	store      *Store
+	provision  *provisionClient
+	interfaces map[string]iface.Interface
```
rodneyosodo, line 21: `interfaces` is mutated from `OpenIface`, `CloseIface`, `Remove`, and `Close`, and read from `ReadIface`/`WriteIface`. With the scheduler goroutine calling `OpenIface`/`ReadIface`/`CloseIface` while HTTP handlers can call `Remove`/`OpenIface` concurrently, this is a plain data race on a Go map — the runtime will throw `concurrent map writes` under load. Wrap access in a `sync.Mutex` (or `sync.Map`), and snapshot the iface pointer under the lock before doing the actual Read/Write/Close I/O outside it.

pkg/devicemgr/manager.go
```diff
@@ -0,0 +1,157 @@
...
+		ifaceCfg:   ifaceCfg,
+	}, nil
+}
+
+// Close releases all open interfaces and the store.
+func (m *Manager) Close() error {
+	for id, ifc := range m.interfaces {
+		_ = ifc.Close()
+		delete(m.interfaces, id)
+	}
```
rodneyosodo: Errors from `ifc.Close()` are dropped on shutdown. At minimum log them so a stuck interface is visible; better, collect with `errors.Join` and return.

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,119 @@
...
+// The four steps mirror what provision.sh does in Steps 2–5.
+func (p *provisionClient) Provision(name, externalID, externalKey string) (Device, error) {
+	if p.cfg.ClientsURL == "" {
+		return Device{}, fmt.Errorf("provision clients URL not configured")
+	}
+	if p.cfg.Token == "" {
+		return Device{}, fmt.Errorf("provision token (PAT) not configured")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
rodneyosodo: `context.Background()` ignores the caller's cancellation entirely. If the agent is shutting down or the HTTP request is cancelled, this provision flow keeps running for the full 30 s and will happily create a half-rolled-back client on a remote service. Plumb the caller's context in (e.g. take `ctx` as the first arg of `Provision`) and only apply the timeout via `context.WithTimeout(ctx, …)`.

pkg/devicemgr/provision.go
```diff
@@ -0,0 +1,119 @@
...
+	// Step 3: Connect client → channel (publish + subscribe).
+	sdkErr = p.sdk.Connect(ctx, mgSDK.Connection{
+		ClientIDs:  []string{client.ID},
+		ChannelIDs: []string{channel.ID},
+		Types:      []string{"publish", "subscribe"},
+	}, p.cfg.DomainID, pat)
+	if sdkErr != nil {
+		_ = p.sdk.DeleteClient(ctx, client.ID, p.cfg.DomainID, pat)
+		_ = p.sdk.DeleteChannel(ctx, channel.ID, p.cfg.DomainID, pat)
+		return Device{}, fmt.Errorf("connect client %s to channel %s: %s", client.ID, channel.ID, sdkErr)
```
rodneyosodo, line 95: Rollback uses the same `ctx` that just timed out / errored. If the parent context is dead, these deletes silently fail and you leak a Magistrala client + channel on the server. Use `context.WithoutCancel(ctx)` (or a fresh short-timeout context) for rollback so cleanup runs even when the main flow has been cancelled. Also log the rollback error rather than `_ =`-ing it — a leaked client is a useful signal.

pkg/devicemgr/store.go
```diff
@@ -0,0 +1,137 @@
...
+	return devices, err
+}
+
+// FindByAddr returns the first active device matching the given interface type
+// and address. Returns an error if no match is found.
+func (s *Store) FindByAddr(ifaceType iface.InterfaceType, addr string) (Device, error) {
+	var found Device
+	var ok bool
+	err := s.db.View(func(tx *bolt.Tx) error {
+		return tx.Bucket(devicesBucket).ForEach(func(_, v []byte) error {
```
rodneyosodo, line 98: `ForEach` keeps iterating over every device after a match is found. For small N it doesn't matter, but more importantly: if two active devices ever share `(InterfaceType, InterfaceAddr)`, the returned device depends on undefined bbolt iteration order. Either return a sentinel error from the callback to break early once `ok = true`, or document that ambiguous addresses pick an arbitrary device.

pkg/iface/iface.go
```diff
@@ -0,0 +1,88 @@
...
+		c := cfg.Modbus
+		c.Mode = modbus.ModeTCP
+		c.Address = addr
+		return modbus.New(c), nil
+	case InterfaceI2C:
+		c := cfg.I2C
+		c.Bus = addr
+		return i2c.New(c), nil
+	default:
+		return nil, fmt.Errorf("unsupported interface type: %s", ifaceType)
```
rodneyosodo, line 90: `InterfaceBLE` and `InterfaceZigbee` are first-class enum values and the UI exposes them in the dropdown, but `New()` falls through to `"unsupported interface type"`. A user provisioning a BLE sensor will succeed end-to-end (client/channel created on the server) and only get an error later when the scheduler tries to open the iface — possibly long after they've moved on. Either implement stubs that surface a clear `not implemented` error, or remove these values from the enum and the UI until they're real.

pkg/iface/i2c/i2c.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	n, err := d.f.Write(data)
+	if err != nil {
+		return n, fmt.Errorf("i2c: write: %w", err)
+	}
+	return n, nil
+}
+
+// ensure unsafe is used (required by the ioctl call path via unix package).
+var _ = unsafe.Sizeof(0)
```
rodneyosodo: This trailing `var _ = unsafe.Sizeof(0)` exists only to silence an unused-import warning, but nothing in this file actually needs `unsafe` — `unix.IoctlSetInt` doesn't take any unsafe pointers. Drop both the `unsafe` import and this line.

pkg/iface/i2c/i2c_unsupported.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// New returns an I2C stub that always errors.
+func New(cfg Config) *I2C {
+	return &I2C{cfg: cfg}
+}
+
+func (d *I2C) Open() error                { return fmt.Errorf("i2c: not supported on this platform") }
+func (d *I2C) Close() error               { return nil }
+func (d *I2C) Read(buf []byte) (int, error)  { return 0, fmt.Errorf("i2c: not supported on this platform") }
```
rodneyosodo: Inconsistent formatting in this block — three different parenthesised-arg-list widths. `gofmt` will normalise this; consider running it.

scheduler.go
```diff
@@ -0,0 +1,212 @@
...
+	"sync"
+	"time"
+
+	"github.com/absmach/agent/pkg/devicemgr"
+	mqtt "github.com/eclipse/paho.mqtt.golang"
+)
+
+const (
+	schedulerReadSize  = 4096
+	schedulerReconnect = 5 * time.Second
```
rodneyosodo: Reconnect delay is a hardcoded 5 s with no backoff and no cap on attempts. If a device is permanently misconfigured (wrong creds, missing physical bus), this goroutine retries forever at fixed cadence, spamming the broker and the logs. Suggest exponential backoff (e.g. 5 s → 30 s → 5 min, capped) and probably structured visibility for "device has been failing for X minutes".

scheduler.go
```diff
@@ -0,0 +1,212 @@
...
+		domainID: domainID,
+		logger:   logger,
+		cancels:  make(map[string]context.CancelFunc),
+	}
+}
+
+// Start stores the parent context and launches goroutines for all active devices
+// already in the registry (devices persisted from a prior run).
+func (s *Scheduler) Start(ctx context.Context) error {
+	s.ctx = ctx
```
rodneyosodo: `s.ctx = ctx` is an unsynchronised write, and `StartDevice` (line 71) reads `s.ctx` without a lock — that's a data race when `AddDevice` runs concurrently with `Start`. Worse, if any caller invokes `StartDevice` before `Start` (e.g. tests, or any future refactor), `s.ctx` is the zero value and `context.WithCancel(nil)` will panic. Pass the context into `newScheduler` and store it in the struct at construction time, or guard `s.ctx` with the mutex and check for nil.

scheduler.go
```diff
@@ -0,0 +1,212 @@
...
+			log.Warn("Device MQTT connect failed", slog.Any("error", err))
+			if !sleepCtx(ctx, schedulerReconnect) {
+				return
+			}
+			continue
+		}
+
+		if err := s.devices.OpenIface(d.ID); err != nil {
+			log.Warn("Device interface open failed", slog.Any("error", err))
+			mqttClient.Disconnect(250)
```
rodneyosodo: On a failed `OpenIface` you `Disconnect(250)` and loop. But `Disconnect(250)` blocks up to 250 ms in the MQTT client; combined with the 5 s `sleepCtx` and an unbounded loop, a single bad device hold-up still gates other unrelated logging and shutdown paths only loosely. Consider a non-blocking disconnect on failure paths.

scheduler.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+// connectDevice creates a fresh MQTT connection authenticated as the device.
+// Device credentials (ID/Key) are used as the MQTT ClientID/Password.
+// TLS settings are inherited from the gateway's MQTT config (same broker).
+func (s *Scheduler) connectDevice(d devicemgr.Device) (mqtt.Client, error) {
+	opts := mqtt.NewClientOptions().
+		AddBroker(s.mqttCfg.URL).
+		SetClientID(d.ID).
```
rodneyosodo, line 189: Using the device ID as both `ClientID` *and* `Username` is fine on Magistrala, but worth a comment — anyone porting this to a vanilla MQTT broker will assume Username = the device's external identity. Also: if two agents accidentally share a device ID (e.g. snapshot/restore of the BoltDB), brokers will kick each other in a connect storm because of the duplicate ClientID. A short note in the file header would save someone a debugging session.

service.go
```diff
@@ -129,6 +138,33 @@ type Service interface {
...
+
+	// AddDevice provisions and registers a new downstream device.
+	AddDevice(name, extID, extKey, ifaceType, ifaceAddr string) (devicemgr.Device, error)
+
+	// RemoveDevice removes a downstream device by ID.
+	RemoveDevice(id string) error
+
+	// MarkDeviceSeen records a live heartbeat for a downstream device.
+	MarkDeviceSeen(id string) error
+}
```
rodneyosodo, line 181: The `Service` interface now exposes 6 new device methods plus `OTAStatus`. Consider extracting a `DeviceService` sub-interface and composing it: `type Service interface { CoreService; DeviceService; OTAService }`. It keeps the mock surface and the new middleware files cleaner, and makes future additions (e.g. `UpdateDevice`) less of a churn.

service.go
```diff
@@ -176,7 +225,7 @@ func (a *agent) changeDir(cmdArr []string) (string, error) {
 		return "sh: cd: " + target + ": No such file or directory", nil
 	}
 	a.workDir = target
-	return "(no output)", nil
+	return "", nil
```
rodneyosodo, line 256: Returning `""` instead of `"(no output)"` is a user-visible behaviour change for anyone scripting against `Execute` or `cd`. The PR description doesn't list it under "breaking changes". Either call it out explicitly or keep the sentinel — and if the goal is just to stop the UI from rendering `(no output)`, fix that on the UI side.

service.go
```diff
@@ -750,3 +860,141 @@ func (a *agent) getTopic(topic string) (t string) {
...
+//	open,<device_id>                             → open physical interface
+//	close,<device_id>                            → close physical interface
+//	read,<device_id>,<n_bytes>                   → read n bytes, reply as hex
+//	write,<device_id>,<hex_data>                 → write hex bytes to interface
+func (a *agent) DeviceManager(uuid, cmdStr string) error {
+	if a.devices == nil {
+		return errDeviceManagerDisabled
+	}
+
+	args := strings.Split(strings.TrimSpace(cmdStr), ",")
```
rodneyosodo, line 1092: `DeviceManager` splits `cmdStr` on `,`. A device `name` containing a comma in `add,name,with,comma,ext_id,...` will silently desync the arg positions and produce a nonsense provisioning call (or worse, succeed with the wrong fields). Either validate `name` (and friends) server-side, or move to a JSON payload.

scripts/provision.sh
```diff
@@ -286,7 +315,11 @@ payload = {
     "profile_id": os.environ["PROFILE_ID"],
     "status": "enabled",
     "render_context": {
-        "mqtt_url": os.environ["MG_AGENT_MQTT_URL"],
+        "mqtt_url":          os.environ["MG_AGENT_MQTT_URL"],
+        "clients_url":       os.environ["MG_AGENT_CLIENTS_URL"],
+        "channels_url":      os.environ["MG_AGENT_CHANNELS_URL"],
+        "provision_token":   os.environ["MG_PAT"],
```
rodneyosodo, line 328: Embedding `MG_PAT` (the operator's Personal Access Token) into the bootstrap `render_context` means every agent on every gateway holds an operator-equivalent credential, persisted to disk via Magistrala's bootstrap config. If a gateway is ever physically compromised, the attacker walks away with full Magistrala authority. Worth at least a comment here calling that out, and ideally moving toward per-device scoped tokens before this ships to anyone with real infra.

cmd/main.go
```diff
@@ -125,7 +132,40 @@ func main() {
...
+		RulesEngineURL: rulesEngineURL,
+		Token:          provisionToken,
+		DomainID:       cfg.DomainID,
+	}, iface.Config{})
+	if err != nil {
+		logger.Error("Failed to open device store", slog.Any("error", err))
+		exitCode = 1
+		return
+	}
+	defer devices.Close()
```
rodneyosodo: `defer devices.Close()` — but `Close()` only logs errors silently (see manager.go:42). On shutdown a stuck interface could swallow data without anyone knowing. Consider logging the close result here.

ui/src/pages/devices.tsx
```diff
@@ -0,0 +1,307 @@
...
+      await load();
+    } catch (e) {
+      setAddError(String(e));
+    } finally {
+      setAdding(false);
+    }
+  }
+
+  async function handleRemove(id: string) {
+    try {
```
rodneyosodo, line 120: `handleRemove` issues the DELETE without any confirmation. Given that removing a device tears down the Magistrala client + channel + rule on the server side (irreversible from the UI), at minimum prompt with a `confirm()` or a modal — a mis-click here is expensive.

ui/src/pages/devices.tsx
```diff
@@ -0,0 +1,307 @@
...
+      const res = await fetch("/api/devices");
+      if (!res.ok) throw new Error(await extractError(res));
+      const data = await res.json();
+      setDevices(data.devices ?? []);
+    } catch (e) {
+      setError(String(e));
+    } finally {
+      setLoading(false);
+    }
+  }
```
rodneyosodo, line 92: `extractError` swallows the JSON parse error and falls through to `HTTP {status}`. If the server returned a plain-text 5xx body (common with some proxies in front), the user gets nothing useful. Consider falling back to `await res.text()` before giving up.

ui/src/pages/ota.tsx
```diff
@@ -0,0 +1,193 @@
...
+      const res = await fetch("/api/ota/status");
+      if (res.ok) setStatus(await res.json());
+    } catch {
+      // network error — keep last known state
+    }
+  }
+
+  useEffect(() => {
+    pollStatus();
+    const id = setInterval(pollStatus, 3000);
```
rodneyosodo: `setInterval(pollStatus, 3000)` runs for the entire lifetime of the page, even when the agent is idle and there's nothing to report. Consider polling only while `status.busy` is true (and once after `submitted` flips), or using SSE — a tab left open will hammer the endpoint indefinitely.

ui/src/pages/ota.tsx
```diff
@@ -0,0 +1,193 @@
...
+    try {
+      const body: Record<string, string | number> = { url };
+      if (sha256.trim()) body.sha256 = sha256.trim();
+      const res = await fetch("/api/ota", {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(body),
+      });
+      if (!res.ok) {
+        const txt = await res.text();
```
rodneyosodo: `const txt = await res.text()` then shown to the user — but the server returns a JSON error body `{"error":"..."}` (see `api/common.go`). The user sees the raw JSON. Parse it the same way `devices.tsx` does (`extractError`) or hoist that helper to a shared module.

ui/src/components/exec-card.tsx
```diff
@@ -51,6 +51,7 @@ export function ExecCard() {
     bottomRef.current?.scrollIntoView({ behavior: "smooth" });
   }, [history, loading]);
 
+
```
rodneyosodo: Stray blank line.

docker/docker-compose.yml
```diff
@@ -19,6 +20,9 @@ services:
     extra_hosts:
       - "host.docker.internal:host-gateway"
       - "bootstrap:host-gateway"
+      - "clients:host-gateway"
```
rodneyosodo: Adding `clients`, `channels`, `rules` to **nodered**'s `extra_hosts` — nodered doesn't talk to those services, so this is dead config that just creates confusion when someone audits the compose file. Remove from the nodered service; keep it only on the agent.

pkg/devicemgr/manager_test.go
```diff
@@ -0,0 +1,519 @@
...
+		{"", iface.InterfaceUnknown},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.input, func(t *testing.T) {
+			got := iface.ParseInterfaceType(tc.input)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
```
rodneyosodo, line 576: `TestParseInterfaceType` here is identical to `TestParseInterfaceType` in `pkg/iface/iface_test.go`. Pick one location (probably the iface package, since `ParseInterfaceType` lives there).

api/transport.go
```diff
@@ -81,6 +81,49 @@ func MakeHandler(svc agent.Service, logger *slog.Logger, stream *logstream.Strea
...
+		EncodeResponse,
+		opts...,
+	).ServeHTTP)
+	r.Get("/devices/{id}", kithttp.NewServer(
+		getDeviceEndpoint(svc),
+		decodeIDFromPath,
+		EncodeResponse,
+		opts...,
+	).ServeHTTP)
+	r.Delete("/devices/{id}", kithttp.NewServer(
```
JeffMboya, line 102: DELETE is not in Access-Control-Allow-Methods (line 36). Browser preflight for this route will be rejected with a CORS error.

service.go
```diff
@@ -724,7 +769,12 @@ func (a *agent) OTA(ctx context.Context, url, sha256hex string, size uint64) err
 	if !a.otaBusy.CompareAndSwap(false, true) {
 		return errors.New("OTA already in progress")
 	}
-	defer a.otaBusy.Store(false)
+	defer func() {
+		a.otaBusy.Store(false)
+	}()
+	a.otaMu.Lock()
+	a.otaLastErr = ""
```
JeffMboya, line 859: otaLastErr is cleared here but never set on failure: the error from ota.Run is returned to the caller but not stored. OTAStatus().LastError will always be empty, making the GET /ota/status endpoint unable to report errors.

api/responses.go
```diff
@@ -117,3 +118,111 @@ func (res nodeRedRes) Headers() map[string]string {
...
+type addDeviceRes struct {
+	devicemgr.Device
+}
+
+func (res addDeviceRes) Code() int {
+	return http.StatusCreated
+}
+
+func (res addDeviceRes) Headers() map[string]string {
+	return map[string]string{}
```
JeffMboya: 201 Created should include a Location header pointing to the new resource (/devices/{id}). Without it, clients must re-list devices to find the ID of what was just created.

scheduler.go
```diff
@@ -0,0 +1,218 @@
...
+
+		token := mqttClient.Publish(topic, s.mqttCfg.QoS, false, data)
+		token.Wait()
+		if err := token.Error(); err != nil {
+			log.Warn("Device publish failed", slog.Any("error", err))
+			return
+		}
+		_ = s.devices.MarkSeen(deviceID)
+	}
+}
```
JeffMboya, line 181: Empty read causes a CPU busy-spin; add a short sleep.

pkg/devicemgr/provision.go
```diff
@@ -4,119 +4,133 @@
...
-		Name: pr.Clients[0].Name,
-	}
-	if d.Name == "" {
-		d.Name = name
-	}
-	if len(pr.Channels) > 0 {
-		d.ChannelID = pr.Channels[0].ID
+	secret := client.Credentials.Secret
+	if secret == "" {
+		secret = externalKey
```
JeffMboya: externalKey is bootstrap credential, not MQTT password.

JeffMboya (approved):
LGTM

PR #46, opened 2026-05-21
AGT-39 - Default to QoS 1 for Command Responses

### LINKED ISSUE #39: Default to QoS 1 for Command Responses
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) publishes command responses and status messages with QoS 1 (at least once delivery) — [`src/transport/mqtt_transport.c`](https://github.com/absmach/aeolus/blob/main/src/transport/mqtt_transport.c).

The Linux agent defaults to QoS 0 for all publishes (configurable via `MG_AGENT_MQTT_QOS`). QoS 0 means fire-and-forget — command responses can be lost in transit with no retry.

**What to change:**

- Default `MG_AGENT_MQTT_QOS` to `1` (at least once)
- Or use QoS 1 specifically for command responses and OTA status publishes (critical messages)
- Keep QoS 0 for heartbeat/telemetry (less critical, next one will arrive anyway)

**Reference:** Aeolus MQTT QoS — [`src/transport/mqtt_transport.c`](https://github.com/absmach/aeolus/blob/main/src/transport/mqtt_transport.c), [`docs/transport.md`](https://github.com/absmach/aeolus/blob/main/docs/transport.md#qos)

JeffMboya:
### What does this do?
Adds `CmdQoS` field to use QoS 1 for command responses and QoS 0 for heartbeat/telemetry.

Resolves #39

- New `MG_AGENT_MQTT_CMD_QOS` env var (default `1`).
- `MG_AGENT_MQTT_QOS` now applies only to heartbeat/telemetry.

Yes. Updated existing tests to assert correct QoS per publish path.

No docs changes needed; env var name is self-explanatory.

### Notes
Matches Aeolus (Zephyr) behaviour: QoS 1 for responses, QoS 0 for heartbeats.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ configs/config.toml
 ui/node_modules/
 ui/dist/*
 !ui/dist/.gitkeep
+!ui/dist/assets
+ui/dist/assets/*
+!ui/dist/assets/placeholder
 ui/.env.local
 ui/tsconfig.tsbuildinfo
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,6 @@ configs/config.toml
 ui/node_modules/
 ui/dist/*
 !ui/dist/.gitkeep
-!ui/dist/assets
-ui/dist/assets/*
-!ui/dist/assets/placeholder
 ui/.env.local
 ui/tsconfig.tsbuildinfo
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,6 +45,7 @@ type config struct {
 	MqttMTLS             string `env:"MG_AGENT_MQTT_MTLS"                     envDefault:"false"`
 	MqttCA               string `env:"MG_AGENT_MQTT_CA"                       envDefault:"ca.crt"`
 	MqttQoS              string `env:"MG_AGENT_MQTT_QOS"                      envDefault:"0"`
+	MqttCmdQoS           string `env:"MG_AGENT_MQTT_CMD_QOS"                  envDefault:"1"`
 	MqttRetain           string `env:"MG_AGENT_MQTT_RETAIN"                   envDefault:"false"`
 	MqttCert             string `env:"MG_AGENT_MQTT_CLIENT_CERT"              envDefault:"client.cert"`
 	MqttPrivateKey       string `env:"MG_AGENT_MQTT_CLIENT_KEY"               envDefault:"client.key"`
@@ -311,6 +312,11 @@ func loadEnvConfig(cfg config) (agent.Config, error) {
 		qos = 0
 	}
 
+	cmdQoS, err := strconv.Atoi(cfg.MqttCmdQoS)
+	if err != nil {
+		cmdQoS = 1
+	}
+
 	retain, err := strconv.ParseBool(cfg.MqttRetain)
 	if err != nil {
 		retain = false
@@ -324,6 +330,7 @@ func loadEnvConfig(cfg config) (agent.Config, error) {
 		PrivKeyPath: cfg.MqttPrivateKey,
 		SkipTLSVer:  skipTLSVer,
 		QoS:         byte(qos),
+		CmdQoS:      byte(cmdQoS),
 		Retain:      retain,
 	}
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -307,13 +307,13 @@ func loadEnvConfig(cfg config) (agent.Config, error) {
 		skipTLSVer = true
 	}
 
-	qos, err := strconv.Atoi(cfg.MqttQoS)
-	if err != nil {
+	qos, err := strconv.ParseUint(cfg.MqttQoS, 10, 8)
+	if err != nil || qos > 2 {
 		qos = 0
 	}
 
-	cmdQoS, err := strconv.Atoi(cfg.MqttCmdQoS)
-	if err != nil {
+	cmdQoS, err := strconv.ParseUint(cfg.MqttCmdQoS, 10, 8)
+	if err != nil || cmdQoS > 2 {
 		cmdQoS = 1
 	}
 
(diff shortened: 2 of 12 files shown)


cmd/main.go
```diff
@@ -42,6 +42,7 @@ type config struct {
 	MqttMTLS             string `env:"MG_AGENT_MQTT_MTLS" envDefault:"false"`
 	MqttCA               string `env:"MG_AGENT_MQTT_CA" envDefault:"ca.crt"`
 	MqttQoS              string `env:"MG_AGENT_MQTT_QOS" envDefault:"0"`
+	MqttCmdQoS           string `env:"MG_AGENT_MQTT_CMD_QOS" envDefault:"1"`
 	MqttRetain           string `env:"MG_AGENT_MQTT_RETAIN" envDefault:"false"`
 	MqttCert             string `env:"MG_AGENT_MQTT_CLIENT_CERT" envDefault:"client.cert"`
```
rodneyosodo: Both `QoS` and `CmdQoS` are parsed from strings without validation against the valid MQTT QoS range (0-2). An out-of-range value (e.g. `3`) would silently be passed to the MQTT client, which may reject the publish or behave unpredictably. This is consistent with the pre-existing `QoS` handling, so not blocking — but worth considering a shared validation helper for both fields.

service.go
```diff
@@ -628,12 +628,19 @@ func (a *agent) Services() []Info {
 }
```
rodneyosodo, line 822: The refactoring from `Publish` to `publishCmd`/`publish` is clean. One minor note: `Ping()` now uses `publishCmd` (QoS 1 by default), whereas the periodic `selfHeartbeat` uses `Publish` (QoS 0). This is a reasonable design choice — on-demand ping via command should ensure delivery (QoS 1), while periodic heartbeats are fire-and-forget (QoS 0). Just confirming this is intentional.

service_test.go
```diff
@@ -41,7 +41,8 @@ func testConfig() agent.Config {
 			Password:   "client-secret",
 			SkipTLSVer: true,
 			Retain:     true,
-			QoS:        1,
```
rodneyosodo, line 51: Changing the test config from `QoS: 1` to `QoS: 0` while adding `CmdQoS: 1` is correct — it now matches the production defaults (`MG_AGENT_MQTT_QOS=0`, `MG_AGENT_MQTT_CMD_QOS=1`). This also exposed that the heartbeat mock in `newService` was previously matching `mock.Anything` for QoS, and now correctly asserts `cfg.MQTT.QoS` (byte 0). Good improvement to test precision.

service_test.go
```diff
@@ -69,11 +70,11 @@ func newService(t *testing.T, cfg agent.Config) (agent.Service, *agentmocks.MQTT
 	return svc, mqttClient, nodeRed, err
 }
 
-func expectMQTTPublish(t *testing.T, mqttClient *agentmocks.MQTTClient, topic string, err error) *mock.Call {
+func expectMQTTPublish(t *testing.T, mqttClient *agentmocks.MQTTClient, topic string, qos byte, err error) *mock.Call {
 	token := agentmocks.NewMQTTToken(t)
 	token.On("Wait").Return(true).Once()
 	token.On("Error").Return(err).Once()
-	return mqttClient.On("Publish", topic, byte(1), true, mock.Anything).Return(token).Once()
```
rodneyosodo, line 111: The new `TestPing` correctly asserts `byte(1)` for the QoS parameter, verifying that `Ping` uses `CmdQoS`. Consider also adding a test case for SenML encoding failure (e.g. by injecting a malformed message) to cover the error path before `publishCmd` is called, though this may be low value given the simplicity of the Ping payload.

cmd/main.go
rodneyosodo: Also, why not parse them as uint8?

ui/dist/assets/placeholder
rodneyosodo, line 1: Use .gitkeep rather same as https://github.com/absmach/agent/pull/59/changes

pkg/ota/ota_test.go
```diff
@@ -67,24 +67,24 @@ func TestTriggerFromRecords(t *testing.T) {
...
 		{
 			desc:    "url only",
-			records: []senml.Record{{Name: "url", StringValue: &urlVal}},
+			records: []senml.Record{{Name: fieldURL, StringValue: &urlVal}},
 			url:     urlVal,
 		},
 		{
 			desc: "url with hash",
 			records: []senml.Record{
-				{Name: "url", StringValue: &urlVal},
```
rodneyosodo, line 76: Ignore `goconst` for *_test.go same as  https://github.com/absmach/agent/pull/59/changes

rodneyosodo (changes requested):

.gitignore
```diff
@@ -6,9 +6,6 @@ configs/config.toml
 ui/node_modules/
 ui/dist/*
 !ui/dist/.gitkeep
-!ui/dist/assets
-ui/dist/assets/*
```
rodneyosodo, line 10: Revert

ui/embed.go
```diff
@@ -11,7 +11,7 @@ import (
 	"strings"
 )
 
-//go:embed dist/*
+//go:embed all:dist
```
rodneyosodo: Revert

tools/config/.golangci.yaml
```diff
@@ -57,6 +57,8 @@ linters:
       template: |-
         Copyright (c) Abstract Machines
         SPDX-License-Identifier: Apache-2.0
+    goconst:
```
rodneyosodo: This config is already there

rodneyosodo (changes requested):

rodneyosodo (approved):

PR #52, opened 2026-05-22
AGT-26 - Add persistent config store with MQTT get/set

### LINKED ISSUE #26: Persistent Config Store with Remote Get/Set
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) has an NVS-backed persistent key-value store ([`src/config/config.c`](https://github.com/absmach/aeolus/blob/main/src/config/config.c)) that survives reboots. Runtime config changes are made via MQTT `config set` commands and persist immediately.

The Linux agent currently has no persistent config store — all configuration comes from environment variables and the bootstrap response at startup. There is no way to change settings at runtime.

**What to implement:**

- Local config file (e.g. JSON or TOML) that persists across restarts
- MQTT `config get <key>` — read a config value, respond on `res` topic
- MQTT `config set <key> <value>` — write a config value to persistent store
- On startup, merge: env vars → bootstrap response → persisted overrides

**Reference:** Aeolus config module — [`src/config/config.c`](https://github.com/absmach/aeolus/blob/main/src/config/config.c), [`docs/config.md`](https://github.com/absmach/aeolus/blob/main/docs/config.md)

JeffMboya:
### What does this do?

Adds a file-backed persistent config store. Supports MQTT `get`/`set` commands. Live-updates log level and heartbeat interval.

Resolves #26

`agent.New()` gains two new parameters: `*config.Store`, `*slog.LevelVar`.

Yes. Store and service config tests added.

No docs changes needed.

### Notes

Settable keys: `log_level`, `heartbeat_interval`, `terminal_session_timeout`.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -429,7 +429,7 @@ func initLogger(levelText string) (*slog.Logger, *slog.LevelVar, error) {
 	return slog.New(logHandler), &levelVar, nil
 }
 
-func applyPersistedOverrides(cfg agent.Config, store *pkgconfig.Store) agent.Config {
+func applyPersistedOverrides(cfg agent.Config, store pkgconfig.Store) agent.Config {
 	for key, val := range store.All() {
 		agent.ApplyConfigEntry(&cfg, key, val)
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -116,6 +116,13 @@ func main() {
 		return
 	}
 	cfg = applyPersistedOverrides(cfg, store)
+	// Sync the live log-level variable with any persisted log_level override.
+	if val, ok := store.Get("log_level"); ok {
+		var l slog.Level
+		if err := l.UnmarshalText([]byte(val)); err == nil {
+			levelVar.Set(l)
+		}
+	}
 
 	if err := validateRuntimeConfig(cfg); err != nil {
 		logger.Error("Failed to validate config", slog.Any("error", err))
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/absmach/agent/api"
 	"github.com/absmach/agent/middleware"
 	"github.com/absmach/agent/pkg/bootstrap"
+	pkgconfig "github.com/absmach/agent/pkg/config"
 	"github.com/absmach/agent/pkg/conn"
 	"github.com/absmach/agent/pkg/devicemgr"
 	"github.com/absmach/agent/pkg/iface"
@@ -59,6 +60,7 @@ type config struct {
 	BootstrapRetryDelay  string `env:"MG_AGENT_BOOTSTRAP_RETRY_DELAY_SECONDS" envDefault:"10"`
 	BootstrapSkipTLS     string `env:"MG_AGENT_BOOTSTRAP_SKIP_TLS" envDefault:"false"`
 	DeviceDBPath         string `env:"MG_AGENT_DEVICE_DB_PATH" envDefault:"/var/lib/agent/devices.db"`
+	ConfigPath           string `env:"MG_AGENT_CONFIG_PATH" envDefault:"agent-config.json"`
 }
 
 var (
@@ -107,6 +109,14 @@ func main() {
 		}
 	}
 
+	store, err := pkgconfig.NewStore(c.ConfigPath)
+	if err != nil {
+		logger.Error("Failed to open persistent config store", slog.Any("error", err))
+		exitCode = 1
+		return
+	}
+	cfg = applyPersistedOverrides(cfg, store)
+
 	if err := validateRuntimeConfig(cfg); err != nil {
 		logger.Error("Failed to validate config", slog.Any("error", err))
 		exitCode = 1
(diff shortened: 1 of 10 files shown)


service_test.go
```diff
@@ -54,6 +55,24 @@ func testConfig() agent.Config {
...
 }
 
+func newServiceWithStore(t *testing.T, cfg agent.Config, store cfgstore.Store) (agent.Service, *agentmocks.MQTTClient, *nrmocks.Client, error) {
+	cfg.DomainID = domainID
+	mqttClient := agentmocks.NewMQTTClient(t)
+	nodeRed := nrmocks.NewClient(t)
+
+	hbToken := agentmocks.NewMQTTToken(t)
+	hbToken.On("Wait").Maybe().Return(true)
+	hbToken.On("Error").Maybe().Return(error(nil))
```
rodneyosodo, line 65: Wrong channel — `New()` uses `cfg.Channels.DataChan()` which resolves to `"data-channel"` in the test config, but this mock expects `"ctrl-channel"`. The expectation will never match.

Change to `mqttTopic("data-channel", "services/agent/heartbeat")`.

service.go
```diff
@@ -290,6 +308,68 @@ func (a *agent) ServiceConfig(ctx context.Context, uuid, cmdStr string) error {
...
+		}
+		if a.store == nil {
+			resp = notConfigured
+		} else if val, ok := a.store.Get(cmdArgs[1]); ok {
+			resp = val
+		} else {
+			resp = notFound
+		}
+	case "set":
+		// Use SplitN(3) so values containing commas (e.g. URLs) are preserved.
```
rodneyosodo: `set` parses from the original `cmdStr` (preserving commas in values), while `get` and `reset` use the pre-split `cmdArgs` (spaces stripped). The result is the same for valid inputs, but the inconsistency is confusing. Consider extracting args from `cmdStr` consistently for all branches.

service.go
```diff
@@ -767,15 +855,107 @@ func (a *agent) Shutdown() {
...
+var settableKeys = map[string]bool{
+	keyLogLevel:               true,
+	keyHeartbeatInterval:      true,
+	keyTerminalSessionTimeout: true,
+}
+
+// validateSettableValue returns errInvalidCommand if val is not a valid value for key.
+func validateSettableValue(key, val string) error {
+	switch key {
+	case keyLogLevel:
```
rodneyosodo, line 868: No `default` case. If a key somehow passes the `settableKeys` check without being handled (e.g. a future key is added to `settableKeys` but not to this switch), the function silently returns nil. Add `default: return errInvalidCommand` as defense-in-depth.

rodneyosodo (commented):
See inline comment below.

rodneyosodo (approved):

service.go
```diff
@@ -767,15 +853,109 @@ func (a *agent) Shutdown() {
...
 func (a *agent) getTopic(topic string) (t string) {
-	domainID := a.config.DomainID
+	cfg := a.Config()
+	domainID := cfg.DomainID
 	switch topic {
 	case control:
-		t = fmt.Sprintf("m/%s/c/%s/res", domainID, a.config.Channels.CtrlChan())
+		t = fmt.Sprintf("m/%s/c/%s/res", domainID, cfg.Channels.CtrlChan())
 	case data:
 		t = fmt.Sprintf("m/%s/c/%s/gateway/telemetry", domainID, a.config.Channels.DataChan())
```
nyagamunene: Should we not update here to `cfg.config.Channels`

cmd/main.go
```diff
@@ -59,6 +60,7 @@ type config struct {
 	BootstrapRetryDelay  string `env:"MG_AGENT_BOOTSTRAP_RETRY_DELAY_SECONDS" envDefault:"10"`
 	BootstrapSkipTLS     string `env:"MG_AGENT_BOOTSTRAP_SKIP_TLS" envDefault:"false"`
 	DeviceDBPath         string `env:"MG_AGENT_DEVICE_DB_PATH" envDefault:"/var/lib/agent/devices.db"`
+	ConfigPath           string `env:"MG_AGENT_CONFIG_PATH" envDefault:"agent-config.json"`
```
nyagamunene, line 63: Should we update the docker compose and .env files with this variable. as of now there is no way to set this to a different path

service_test.go
JeffMboya: Updated mock to use data-channel with gateway/heartbeat topic.

service.go
JeffMboya: All branches now parse from cmdStr consistently using TrimSpace.

service.go
JeffMboya: Added default: return errInvalidCommand to validateSettableValue.

service.go
JeffMboya: Fixed — getTopic now uses cfg.Channels consistently throughout.

cmd/main.go
JeffMboya: Added MG_AGENT_CONFIG_PATH to docker/.env and docker-compose.yml.

nyagamunene (approved):

rodneyosodo (approved):

PR #54, opened 2026-05-25
AG-25 - Enrich Heartbeat Payload

### LINKED ISSUE #25: Rich Heartbeat Payload
The Linux agent heartbeat currently publishes only a single `service_type` field:

```json
[{"bn":"agent:","n":"service_type","vs":"agent"}]
```

The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) publishes a rich health pack every heartbeat interval (see [`src/management/health.c`](https://github.com/absmach/aeolus/blob/main/src/management/health.c)):

```json
[
  {"n":"heartbeat","vb":true},
  {"n":"fw_version","vs":"4.2.0"},
  {"n":"uptime","u":"s","v":3600},
  {"n":"heap_free","u":"By","v":12384},
  {"n":"devices","u":"count","v":3},
  {"n":"connected","vb":true}
]
```

The Linux agent should publish the same fields so both agents have a unified heartbeat schema that the cloud side can consume identically.

**Fields to add to `selfHeartbeat()` in `service.go`:**

| Field | Source |
|---|---|
| `heartbeat` | Always `true` |
| `fw_version` / `agent_version` | Build-time version string |
| `uptime` | `time.Since(startTime).Seconds()` |
| `heap_free` | `runtime/metrics` or `runtime.ReadMemStats` |
| `connected` | MQTT client connection state |

**Reference:** Aeolus implementation — [`src/management/health.c`](https://github.com/absmach/aeolus/blob/main/src/management/health.c#L98-L160)

nyagamunene:
Resolves #25 

### Notes

### DIFF
--- a/pkg/devicemgr/store.go
+++ b/pkg/devicemgr/store.go
@@ -75,7 +75,11 @@ func (s *Store) Get(id string) (Device, error) {
 func (s *Store) Count() (int, error) {
 	var n int
 	err := s.db.View(func(tx *bolt.Tx) error {
-		n = tx.Bucket(devicesBucket).Stats().KeyN
+		b := tx.Bucket(devicesBucket)
+		if b == nil {
+			return fmt.Errorf("bucket %q not found", devicesBucket)
+		}
+		n = b.Stats().KeyN
 		return nil
 	})
 	return n, err
--- a/pkg/devicemgr/store.go
+++ b/pkg/devicemgr/store.go
@@ -71,6 +71,16 @@ func (s *Store) Get(id string) (Device, error) {
 	return d, err
 }
 
+// Count returns the number of stored devices without loading the records.
+func (s *Store) Count() (int, error) {
+	var n int
+	err := s.db.View(func(tx *bolt.Tx) error {
+		n = tx.Bucket(devicesBucket).Stats().KeyN
+		return nil
+	})
+	return n, err
+}
+
 // List returns all stored devices.
 func (s *Store) List() ([]Device, error) {
 	var devices []Device
--- a/service.go
+++ b/service.go
@@ -13,7 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
-	"runtime"
+	"runtime/metrics"
 	"sort"
 	"strconv"
 	"strings"
@@ -840,21 +840,24 @@ func (a *agent) selfHeartbeat(ctx context.Context, topic string, interval time.D
 }
 
 func (a *agent) selfHeartbeatPayload() ([]byte, error) {
-	var mem runtime.MemStats
-	runtime.ReadMemStats(&mem)
+	samples := []metrics.Sample{
+		{Name: "/memory/classes/heap/free:bytes"},
+	}
+
+	metrics.Read(samples)
 
 	heapFree := uint64(0)
-	if mem.HeapSys > mem.HeapAlloc {
-		heapFree = mem.HeapSys - mem.HeapAlloc
+	if len(samples) > 0 {
+		heapFree = samples[0].Value.Uint64()
 	}
 
 	deviceCount := 0
 	if a.devices != nil {
-		devices, err := a.devices.List()
+		n, err := a.devices.Count()
 		if err != nil {
-			a.logger.Warn("failed to list devices for self-heartbeat", slog.Any("error", err))
+			a.logger.Warn("failed to count devices for self-heartbeat", slog.Any("error", err))
 		} else {
-			deviceCount = len(devices)
+			deviceCount = n
 		}
 	}
 
--- a/service.go
+++ b/service.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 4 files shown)


service_test.go
```diff
@@ -88,6 +90,77 @@ func expectMQTTPublish(t *testing.T, mqttClient *agentmocks.MQTTClient, topic st
...
+	token.On("Wait").Return(true).Once()
+	token.On("Error").Run(func(_ mock.Arguments) {
+		close(published)
+	}).Return(error(nil)).Once()
+
+	mqttClient.On("Publish", mqttTopic("ctrl-channel", "services/agent/heartbeat"), byte(1), false, mock.MatchedBy(func(payload interface{}) bool {
+		return richHeartbeatPayload(t, payload)
+	})).Return(token).Once()
+
+	ctx, cancel := context.WithCancel(context.Background())
```
rodneyosodo, line 133: Wrong heartbeat topic — `agent.New()` publishes to `m/%s/c/%s/gateway/heartbeat` using `cfg.Channels.DataChan()` (resolves to `"data-channel"`). This mock expects a `ctrl-channel` topic with a `services/agent/heartbeat` suffix that doesn't match.

Fix: `mqttTopic("data-channel", "gateway/heartbeat")`

service.go
```diff
@@ -693,6 +696,45 @@ func (a *agent) selfHeartbeat(ctx context.Context, topic string, interval time.D
 	}
 }
 
+func (a *agent) selfHeartbeatPayload() ([]byte, error) {
+	var mem runtime.MemStats
+	runtime.ReadMemStats(&mem)
```
rodneyosodo: `runtime.ReadMemStats` requires a stop-the-world pause. For typical heartbeat intervals (>= 30s) this is negligible, but at very low intervals (e.g. 1s) it could add up. Consider using `debug.ReadGCStats` or the `runtime/metrics` API if this becomes a concern.

service.go
```diff
@@ -693,6 +696,45 @@ func (a *agent) selfHeartbeat(ctx context.Context, topic string, interval time.D
...
+
+	heapFree := uint64(0)
+	if mem.HeapSys > mem.HeapAlloc {
+		heapFree = mem.HeapSys - mem.HeapAlloc
+	}
+
+	deviceCount := 0
+	if a.devices != nil {
+		devices, err := a.devices.List()
```
rodneyosodo: `a.devices.List()` is O(n) just to get `len(devices)`. If `devicemgr.Manager` supports a `Count()` method, prefer that. Acceptable for now but worth noting for scalability.

rodneyosodo (commented):
## Bug: `TestSelfHeartbeatPublishesRichPayload` expects wrong topic

The mock expects `mqttTopic("ctrl-channel", "services/agent/heartbeat")` but `agent.New()` constructs the heartbeat topic as `m/%s/c/%s/gateway/heartbeat` using `cfg.Channels.DataChan()`. With the test config, `DataChan()` returns `"data-channel"`, so the actual topic is `m/.../c/**data-channel**/gateway/**heartbeat**` — neither the channel nor the suffix match.

The heartbeat goroutine calls `publish()` immediately on startup, so this will fail instantly with an unexpected mock call.

**Note:** The same pre-existing mismatch exists in the `newService` helper (visible in the diff context around line 71), but this test is new code that copies the broken pattern.

pkg/devicemgr/store.go
```diff
@@ -70,6 +70,16 @@ func (s *Store) Get(id string) (Device, error) {
 	return d, err
 }
 
+// Count returns the number of stored devices without loading the records.
+func (s *Store) Count() (int, error) {
+	var n int
+	err := s.db.View(func(tx *bolt.Tx) error {
+		n = tx.Bucket(devicesBucket).Stats().KeyN
```
rodneyosodo: **Potential nil pointer dereference** — `tx.Bucket(devicesBucket)` returns `nil` if the bucket doesn't exist (corrupted DB, etc.), and calling `.Stats()` on a nil `*bolt.Bucket` will panic.

Recommend adding a nil check:

```go
b := tx.Bucket(devicesBucket)
if b == nil {
    return fmt.Errorf("bucket %q not found", devicesBucket)
}
n = b.Stats().KeyN
```

Same vulnerability exists in `Save`, `Get`, `Delete`, etc., so this follows the existing non-defensive pattern — but since this is new code, it's worth making it robust.

rodneyosodo (commented):
## Review

Overall the changes look solid. A few observations below.

service.go
```diff
@@ -693,6 +697,48 @@ func (a *agent) selfHeartbeat(ctx context.Context, topic string, interval time.D
 	}
 }
 
+func (a *agent) selfHeartbeatPayload() ([]byte, error) {
+	samples := []metrics.Sample{
```
rodneyosodo: **Minor:** `samples` is allocated as a new slice on every heartbeat tick. At typical heartbeat intervals (>=30s) this is irrelevant, but if the interval is ever very short (e.g. 1s) it adds unnecessary GC pressure. Consider hoisting it to a package-level or struct-level variable if performance becomes a concern.

service_test.go
```diff
@@ -88,6 +90,77 @@ func expectMQTTPublish(t *testing.T, mqttClient *agentmocks.MQTTClient, topic st
...
+	mqttClient.On("IsConnected").Return(true).Maybe()
+
+	published := make(chan struct{})
+	token := agentmocks.NewMQTTToken(t)
+	token.On("Wait").Return(true).Once()
+	token.On("Error").Run(func(_ mock.Arguments) {
+		close(published)
+	}).Return(error(nil)).Once()
+
+	mqttClient.On("Publish", mqttTopic("data-channel", "gateway/heartbeat"), byte(1), false, mock.MatchedBy(func(payload interface{}) bool {
```
rodneyosodo: **Minor:** The QoS value `byte(1)` is hardcoded here and matches the default from `testConfig()`. If that default ever changes, this test will silently break. Consider using `cfg.MQTT.QoS` instead to keep the test in sync with config automatically.

rodneyosodo (approved):

PR #59, opened 2026-05-26
AG - 32 - Add disk cache and invalidation for bootstrap responses

### LINKED ISSUE #32: Cache Bootstrap Response to Avoid Re-fetch on Restart
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) caches the bootstrap response in NVS ([`src/bootstrap/bootstrap.c`](https://github.com/absmach/aeolus/blob/main/src/bootstrap/bootstrap.c)). On subsequent boots, `bootstrap_load_nvs()` returns the cached data directly — no HTTP fetch is performed. Bootstrap is only re-fetched when the cache is empty or `bs_valid=0`.

The Linux agent re-fetches bootstrap on every startup. This adds latency and creates a hard dependency on the bootstrap service being available at boot time.

**What to implement:**

- Cache the rendered bootstrap response to a local file (e.g. `/var/lib/agent/bootstrap.json`)
- On startup, check cache first; if valid and non-empty, use cached config
- If cache is missing/invalid/corrupt, fetch from bootstrap service and update cache
- Allow cache invalidation via `config set bs_valid 0` or manual file deletion
- Log whether config came from cache or fresh fetch

**Reference:** Aeolus `bootstrap_load_nvs()` / `bootstrap_store_nvs()` — [`src/bootstrap/bootstrap.c`](https://github.com/absmach/aeolus/blob/main/src/bootstrap/bootstrap.c), [`docs/bootstrap.md`](https://github.com/absmach/aeolus/blob/main/docs/bootstrap.md)

rodneyosodo:
### What does this do?

Caches the bootstrap response to a local file so the agent doesn't re-fetch from the bootstrap service on every restart. On startup, if a valid cache exists, it is used directly. If the cache is missing, corrupt, or explicitly invalidated, the agent fetches fresh config from the bootstrap service and writes it to cache.

### Which issue(s) does this fix/relate to?

Resolves #32

- `bootstrap.FetchAgentConfig` signature changed: added `forceFetch bool` parameter
- `agent.New` signature changed: added `bootstrapCachePath string` parameter
- Config store is now opened before bootstrap runs (previously opened after)

Yes

No

### Notes

Mirrors the Aeolus `bootstrap_load_nvs()` / `bootstrap_store_nvs()` pattern (`src/bootstrap/bootstrap.c`). Cache writes are atomic (write to `.tmp`, then `os.Rename`).

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -11,6 +11,8 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -28,6 +30,7 @@ type Config struct {
 	RetryDelaySec string
 	Encrypt       string
 	SkipTLS       bool
+	CachePath     string
 }
 
 // renderedContent holds the device configuration rendered by the bootstrap
@@ -72,7 +75,19 @@ type bootstrapResponse struct {
 
 // FetchAgentConfig retrieves device configuration from the bootstrap service and
 // overlays the returned credentials and channel IDs onto agentCfg.
-func FetchAgentConfig(cfg Config, agentCfg agent.Config, logger *slog.Logger) (agent.Config, error) {
+// When cfg.CachePath is set and forceFetch is false, it first attempts to load
+// a previously cached response from disk. On cache miss it fetches from the
+// bootstrap service and writes the response to cfg.CachePath.
+func FetchAgentConfig(cfg Config, agentCfg agent.Config, logger *slog.Logger, forceFetch bool) (agent.Config, error) {
+	if !forceFetch && cfg.CachePath != "" {
+		br, err := loadFromCache(cfg.CachePath)
+		if err == nil {
+			logger.Info("Loaded bootstrap config from cache", slog.String("path", cfg.CachePath))
+			return applyBootstrapResponse(agentCfg, br)
+		}
+		logger.Info("Bootstrap cache miss, fetching from service", slog.String("path", cfg.CachePath), slog.Any("cache_error", err))
+	}
+
 	retries, err := strconv.ParseUint(cfg.Retries, 10, 64)
 	if err != nil {
 		return agentCfg, errors.New(fmt.Sprintf("Invalid BOOTSTRAP_RETRIES value: %s", err))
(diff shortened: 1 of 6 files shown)


service.go
```diff
@@ -1106,6 +1114,14 @@ func ApplyConfigEntry(cfg *Config, key, val string) {
 // effect is immediate without requiring a restart.
 func (a *agent) applyLiveUpdate(key, val string) {
 	switch key {
+	case keyBsValid:
+		if val == "0" && a.bootstrapCachePath != "" {
+			if err := os.Remove(a.bootstrapCachePath); err != nil && !os.IsNotExist(err) {
+				a.logger.Warn("Failed to delete bootstrap cache", slog.Any("error", err))
+			} else {
+				a.logger.Info("Bootstrap cache invalidated")
```
JeffMboya, line 1146: Logs 'invalidated' even when file was already absent.

pkg/bootstrap/bootstrap.go
```diff
@@ -72,7 +75,19 @@ type bootstrapResponse struct {
...
+// a previously cached response from disk. On cache miss it fetches from the
+// bootstrap service and writes the response to cfg.CachePath.
+func FetchAgentConfig(cfg Config, agentCfg agent.Config, logger *slog.Logger, forceFetch bool) (agent.Config, error) {
+	if !forceFetch && cfg.CachePath != "" {
+		br, err := loadFromCache(cfg.CachePath)
+		if err == nil {
+			logger.Info("Loaded bootstrap config from cache", slog.String("path", cfg.CachePath))
+			return applyBootstrapResponse(agentCfg, br)
+		}
+		logger.Info("Bootstrap cache miss, fetching from service", slog.String("path", cfg.CachePath), slog.Any("cache_error", err))
```
JeffMboya: Inconsistent key; use "error" like everywhere else.

pkg/bootstrap/bootstrap.go
```diff
@@ -229,3 +252,34 @@ func getConfig(bsID, bsKey, bsSvrURL string, skipTLS bool, logger *slog.Logger)
...
+func storeToCache(path string, br bootstrapResponse) error {
+	data, err := json.MarshalIndent(br, "", "  ")
+	if err != nil {
+		return err
+	}
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return err
+	}
+	tmp := path + ".tmp"
```
JeffMboya, line 283: .tmp orphaned if WriteFile succeeds but Rename fails.

pkg/bootstrap/bootstrap.go
```diff
@@ -229,3 +252,34 @@ func getConfig(bsID, bsKey, bsSvrURL string, skipTLS bool, logger *slog.Logger)
...
+func loadFromCache(path string) (bootstrapResponse, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return bootstrapResponse{}, err
+	}
+	var br bootstrapResponse
+	if err := json.Unmarshal(data, &br); err != nil {
+		return bootstrapResponse{}, err
+	}
+	if br.Content == "" {
```
JeffMboya, line 265: Empty ClientKey/Cert accepted; cache may overwrite live creds.

JeffMboya (approved):
LGTM

PR #61, opened 2026-05-28
AG-33 - Extensible command handler registry

### LINKED ISSUE #33: Extensible Command Handler Registry
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) uses a registerable command handler system ([`src/control/control.c`](https://github.com/absmach/aeolus/blob/main/src/control/control.c)):

```c
control_register_handler("ota", ota_cmd_handler);
```

New commands are added by registering a handler function. The dispatch in `cmd_message_handler()` falls through to `control_dispatch_record()` for any unrecognized record name.

The Linux agent uses a hard-coded `switch` statement in `conn.go` (`handleMsg`). Adding a new command requires modifying the switch.

**What to implement:**

- `CommandHandler` interface or `func(*SenMLRecord) error` type
- `RegisterHandler(name string, handler CommandHandler)` method on Service
- Map-based dispatch in `handleMsg` — look up handler by record name, fall through to default for unknown commands
- Register built-in handlers (`exec`, `config`, `term`, `nodered`, `ping`, `reset`, `ota`) at init time
- Allow external packages to register custom handlers

**Reference:** Aeolus control module — [`src/control/control.c`](https://github.com/absmach/aeolus/blob/main/src/control/control.c)

nyagamunene:
Resolves #33

### Notes

### DIFF
--- a/pkg/conn/conn.go
+++ b/pkg/conn/conn.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
 	"syscall"
 
 	"github.com/absmach/agent"
@@ -38,14 +39,22 @@ const (
 
 var channelPartRegExp = regexp.MustCompile(`^m/([\w\-]+)/c/([\w\-]+)/services(/[^?]*)?(\?.*)?$`)
 
+// CommandHandler processes a single inbound MQTT command.
+// pack contains the full decoded SenML pack; pack.Records[0] is always the
+// command record. Multi-record commands (e.g. OTA) use pack.Records[1:].
+type CommandHandler func(ctx context.Context, pack senml.Pack) error
+
 var _ MqttBroker = (*broker)(nil)
 
 // MqttBroker represents the MQTT broker.
 type MqttBroker interface {
-	// Subscribes to given topic and receives events.
+	// Subscribe subscribes to given topic and receives events.
 	Subscribe(ctx context.Context) error
 	// Resubscribe re-runs topic subscriptions after a reconnect.
 	Resubscribe()
+	// RegisterHandler registers a CommandHandler for the given command name.
+	// Calling RegisterHandler with an existing name replaces the previous handler.
+	RegisterHandler(name string, h CommandHandler)
 }
 
 type broker struct {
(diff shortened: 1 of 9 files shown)


pkg/conn/conn.go
```diff
@@ -55,16 +64,122 @@ type broker struct {
...
+		b.logger.Warn("RegisterHandler called with nil handler", slog.String("name", name))
+		return
+	}
+	b.mu.Lock()
+	b.handlers[name] = h
+	b.mu.Unlock()
+}
+
+// registerBuiltins wires the built-in command set into the handler registry.
+func (b *broker) registerBuiltins() {
```
JeffMboya, line 99: Writes to `handlers` without mutex, inconsistent with `RegisterHandler`; safe only during construction.

pkg/conn/conn.go
```diff
@@ -55,16 +64,122 @@ type broker struct {
...
+	}
+
+	b.handlers[reset] = func(ctx context.Context, pack senml.Pack) error {
+		uuid, _ := extractCmd(pack)
+		log.Info("Reset command received, performing graceful shutdown", slog.String("uuid", uuid))
+		svc.Shutdown()
+		if err := syscall.Exec(os.Args[0], os.Args, os.Environ()); err != nil {
+			log.Error("Reset failed", slog.Any("error", err))
+		}
+		return nil
```
JeffMboya, line 154: Returns `nil` on `Exec` failure; the reset silently aborts with no recovery.

pkg/conn/conn.go
```diff
@@ -55,16 +64,122 @@ type broker struct {
...
+
+	b.handlers[otaCmd] = func(ctx context.Context, pack senml.Pack) error {
+		uuid, _ := extractCmd(pack)
+		trigger, err := ota.TriggerFromRecords(pack.Records[1:])
+		if err != nil {
+			return err
+		}
+		log.Info("OTA command", slog.String("uuid", uuid), slog.String("url", trigger.URL))
+		go func() {
+			if err := svc.OTA(ctx, trigger.URL, trigger.SHA256Hex, trigger.Size); err != nil {
```
JeffMboya: OTA goroutine inherits cancellable `ctx`; broker shutdown may abort a live firmware update.

JeffMboya (approved):
LGTM

PR #66, opened 2026-06-09
AG - 36 - Add command for secret authorization

### LINKED ISSUE #36: Command Token Authorization for MQTT Commands
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) supports optional token-based authorization for MQTT commands ([`src/agent.c`](https://github.com/absmach/aeolus/blob/main/src/agent.c)):

When `command_secret` is set in NVS, every inbound SenML pack must contain a `{"n":"token","vs":"..."}` record matching the stored secret. Comparison is constant-time to prevent timing attacks. Packs without a valid token are silently rejected.

The Linux agent has no command authorization — any client with publish access to the `req` topic can issue commands.

**What to implement:**

- Optional `MG_AGENT_COMMAND_SECRET` env var
- If set, require a `token` record in every command SenML pack
- Constant-time string comparison (e.g. `crypto/subtle.ConstantTimeCompare`)
- Log rejected commands, publish no response (cloud-side uses timeout to detect)
- Allow `command_secret` to be set via config command (once persistent config exists)

**Reference:** Aeolus `authorize_command()` — [`src/agent.c`](https://github.com/absmach/aeolus/blob/main/src/agent.c#L332-L365), [`docs/control.md`](https://github.com/absmach/aeolus/blob/main/docs/control.md#authorization)

rodneyosodo:
### What does this do?

Implements optional token-based authorisation for inbound MQTT commands. When `MG_AGENT_COMMAND_SECRET` is set (or configured via the `command_secret` config key), every inbound SenML pack must contain a `{"n":"token","vs":"<secret>"}` record matching the stored secret. Comparison uses `crypto/subtle.ConstantTimeCompare` to prevent timing attacks. Packs without a valid token are silently rejected and logged — no response is published, so the cloud-side detects rejection via timeout.

The `command_secret` can also be set/updated at runtime via the MQTT config command (`config,set,command_secret,<value>`), with the change taking effect immediately on the next inbound message.

Resolves #36

- **Behavioral change when `command_secret` is configured**: Any MQTT client publishing to the `req` topic without a valid `token` record will have its commands silently dropped. Existing clients must be updated to include the token in their SenML packs.
- New `command_secret` key added to the settable config allowlist, readable/writable/resettable via `config,get/set/reset,command_secret`.
- The `Config` struct gains a `CommandSecret` field (excluded from JSON serialization).

Yes:
- `pkg/conn/conn_test.go` — 6 test cases covering valid/invalid/missing/empty token scenarios
- `service_test.go` — added test cases for `command_secret` in `TestConfigGetSet` (set/get/reset) and `TestApplyConfigEntry`

No documentation changes in this PR. The following should be documented:
- `MG_AGENT_COMMAND_SECRET` environment variable
- Required `token` SenML record format for authenticated commands
- `command_secret` as a settable config key via MQTT config commands

### Notes

- The broker reads the secret from `svc.Config()` on every message rather than storing a static copy, so runtime updates via `config,set,command_secret,<value>` take effect immediately without restart.
- No live MQTT reconnect is triggered when the secret changes — this is intentional since the secret affects inbound command validation, not the agent's own MQTT connection.
- Follows the same pattern as the Zephyr agent (Aeolus) `authorize_command()` implementation.

### DIFF
--- a/pkg/conn/conn.go
+++ b/pkg/conn/conn.go
@@ -5,6 +5,7 @@ package conn
 
 import (
 	"context"
+	"crypto/subtle"
 	"fmt"
 	"log/slog"
 	"os"
@@ -262,6 +263,14 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 		return
 	}
 
+	commandSecret := b.svc.Config().CommandSecret
+	if commandSecret != "" {
+		if !authorizeCommand(records, commandSecret) {
+			b.logger.Warn("Command rejected: invalid or missing token")
+			return
+		}
+	}
+
 	cmdType := records[0].Name
 
 	b.mu.RLock()
@@ -276,6 +285,18 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 	if err := h(b.ctx, sm); err != nil {
 		b.logger.Warn("command handler failed", slog.String("command", cmdType), slog.Any("error", err))
 	}
+}
+
+// authorizeCommand checks whether the SenML pack contains a valid token record.
+// Returns false if the token is missing or does not match the stored secret.
+// Uses constant-time comparison to prevent timing attacks.
+func authorizeCommand(records []senml.Record, secret string) bool {
+	for _, r := range records {
+		if r.Name == "token" && r.StringValue != nil {
+			return subtle.ConstantTimeCompare([]byte(*r.StringValue), []byte(secret)) == 1
+		}
+	}
+	return false
 }
 
 // extractCmd returns the uuid and string value from the first SenML record.
--- a/pkg/conn/conn.go
+++ b/pkg/conn/conn.go
@@ -263,7 +263,7 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 		return
 	}
 
-	commandSecret := b.svc.Config().CommandSecret
+	commandSecret := b.svc.CommandSecret()
 	if commandSecret != "" {
 		if !authorizeCommand(records, commandSecret) {
 			b.logger.Warn("Command rejected: invalid or missing token")
--- a/pkg/conn/conn_test.go
+++ b/pkg/conn/conn_test.go
@@ -45,12 +45,7 @@ func TestAuthorizeCommand(t *testing.T) {
 			secret:  "my-secret-token",
 			want:    false,
 		},
-		{
-			desc:    "empty secret matches empty token",
-			records: []senml.Record{otherRecord, {Name: "token", StringValue: new("")}},
-			secret:  "",
-			want:    true,
-		},
+
 		{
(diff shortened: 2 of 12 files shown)


service.go
```diff
@@ -52,6 +52,7 @@ const (
```
JeffMboya, line 51: test

service.go
```diff
@@ -1037,6 +1038,7 @@ var settableKeys = map[string]bool{
 	keyLogLevel:               true,
 	keyHeartbeatInterval:      true,
 	keyTerminalSessionTimeout: true,
+	keyCommandSecret:          true,
```
JeffMboya, line 1054: config,get,command_secret returns the stored secret.

service.go
```diff
@@ -1057,8 +1059,8 @@ func validateSettableValue(key, val string) error {
 		if err != nil || d <= 0 {
 			return errInvalidCommand
 		}
-	default:
-		return errInvalidCommand
+	case keyCommandSecret:
+		// Any non-empty string is valid; empty value clears the secret.
 	}
```
JeffMboya, line 1079: Missing default; new keys skip validation silently.

service.go
```diff
@@ -1078,6 +1080,9 @@ func (a *agent) revertToStartup(key string) {
 	case keyTerminalSessionTimeout:
 		a.config.Terminal.SessionTimeout = a.startupConfig.Terminal.SessionTimeout
 		liveVal = a.config.Terminal.SessionTimeout.String()
+	case keyCommandSecret:
+		a.config.CommandSecret = a.startupConfig.CommandSecret
+		liveVal = a.config.CommandSecret
```
JeffMboya, line 1100: liveVal passed to applyLiveUpdate but no case handles it.

pkg/conn/conn.go
```diff
@@ -262,6 +263,14 @@ func (b *broker) handleMsg(msg mqtt.Message) {
 		return
 	}
 
+	commandSecret := b.svc.Config().CommandSecret
```
JeffMboya: Config() snapshot races concurrent secret update.

pkg/conn/conn_test.go
```diff
@@ -0,0 +1,68 @@
...
+			want:    false,
+		},
+		{
+			desc:    "token record with nil string value returns false",
+			records: []senml.Record{otherRecord, {Name: "token"}},
+			secret:  "my-secret-token",
+			want:    false,
+		},
+		{
+			desc:    "empty secret matches empty token",
```
JeffMboya: Unreachable: handleMsg skips auth when secret is empty.

JeffMboya (approved):
LGTM

PR #67, opened 2026-06-10
AG - 40 - Persist bootstrap fields and skip fetch if present

### LINKED ISSUE #40: Skip Bootstrap Fetch When Credentials Already Present
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) skips the bootstrap fetch entirely when credentials are already available ([`src/agent.c`](https://github.com/absmach/aeolus/blob/main/src/agent.c)):

```c
if (g_cfg.domain_id[0] != '\0' && g_cfg.data_channel[0] != '\0' &&
    g_cfg.device_id[0] != '\0') {
    LOG_INF("Bootstrap data already present, skipping");
    return 0;
}
```

The Linux agent always calls `bootstrap.FetchAgentConfig()` when bootstrap env vars are set, even if the runtime config already has all required fields from a previous run.

**What to implement:**

- Before calling bootstrap fetch, check if the runtime config already has all required fields (domain_id, channels, mqtt credentials)
- If all present and non-empty, skip the HTTP fetch
- This works best combined with bootstrap caching (separate issue)

**Reference:** Aeolus `do_bootstrap()` — [`src/agent.c`](https://github.com/absmach/aeolus/blob/main/src/agent.c#L152-L162)

rodneyosodo:
## What does this do?

Skips the bootstrap HTTP fetch when the agent already has valid runtime credentials from a previous run. 

On startup, the agent now checks whether `domain_id`, channels (`ctrl_id` / `data_id`), and MQTT credentials (URL + username/password, or mTLS) are already present before calling `bootstrap.FetchAgentConfig()`. After a successful bootstrap, these fields are persisted to the config store (`agent-config.json`) so they survive restarts. On subsequent starts the agent loads them from the store, detects they're already present, and skips the fetch entirely.

A `bs_valid=0` flag in the config store still forces a fresh fetch regardless.

## Which issue(s) does this PR fix/relate to?

Resolves #40

## List any changes that modify/break current functionality

None. The change is additive — if credentials aren't present the bootstrap flow runs as before. The `bs_valid` flag mechanism is preserved unchanged.

## Have you included tests for your changes?

Yes — added 6 test cases to `TestApplyConfigEntry` covering the new config keys (`domain_id`, `channels_ctrl_id`, `channels_data_id`, `mqtt_url`, `mqtt_username`, `mqtt_password`). Fixed the existing "unknown key is a no-op" test which previously used `mqtt_password` as an unknown key.

## Did you document any new/modified functionality?

No documentation changes — the behavior is transparent to end users.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -129,7 +129,7 @@ func main() {
 				exitCode = 1
 				return
 			}
-			persistBootstrapFields(store, cfg)
+			persistBootstrapFields(store, cfg, logger)
 			if err := store.Set("bs_valid", "1"); err != nil {
 				logger.Warn("Failed to persist bs_valid flag", slog.Any("error", err))
 			}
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -266,6 +273,31 @@ func hasBootstrapConfig(cfg config) bool {
...
+	if cfg.MQTT.URL == "" {
+		return false
+	}
+	if !cfg.MQTT.MTLS && (cfg.MQTT.Username == "" || cfg.MQTT.Password == "") {
+		return false
+	}
+	return true
+}
+
+func persistBootstrapFields(store pkgconfig.Store, cfg agent.Config) {
```
JeffMboya: All store.Set errors silently discarded; persist failures will silently break bootstrap skip on restart.

cmd/main.go
```diff
@@ -266,6 +273,31 @@ func hasBootstrapConfig(cfg config) bool {
...
+	_ = store.Set("domain_id", cfg.DomainID)
+	_ = store.Set("channels_ctrl_id", cfg.Channels.CtrlID)
+	_ = store.Set("channels_data_id", cfg.Channels.DataID)
+	_ = store.Set("mqtt_url", cfg.MQTT.URL)
+	_ = store.Set("mqtt_username", cfg.MQTT.Username)
+	_ = store.Set("mqtt_password", cfg.MQTT.Password)
```
JeffMboya: MQTT password written to agent-config.json in plaintext; worth documenting the security implications.

cmd/main.go
```diff
@@ -266,6 +273,31 @@ func hasBootstrapConfig(cfg config) bool {
...
+	if cfg.DomainID == "" {
+		return false
+	}
+	if cfg.Channels.CtrlID == "" || cfg.Channels.DataID == "" {
+		return false
+	}
+	if cfg.MQTT.URL == "" {
+		return false
+	}
+	if !cfg.MQTT.MTLS && (cfg.MQTT.Username == "" || cfg.MQTT.Password == "") {
```
JeffMboya: mTLS path skips username/password check but doesn't verify cert/key paths are configured.

service.go
```diff
@@ -1223,6 +1229,18 @@ func ApplyConfigEntry(cfg *Config, key, val string) {
...
+		cfg.DomainID = val
+	case keyChannelsCtrlID:
+		cfg.Channels.CtrlID = val
+	case keyChannelsDataID:
+		cfg.Channels.DataID = val
+	case keyMqttURL:
+		cfg.MQTT.URL = val
+	case keyMqttUsername:
+		cfg.MQTT.Username = val
+	case keyMqttPassword:
```
JeffMboya, line 1246: mqtt_password can now be updated at runtime via store; verify command authorization guards this path.

JeffMboya (approved):
LGTM

PR #69, opened 2026-06-10
AG - 38 - Add health supervisor and MQTT checker feature

### LINKED ISSUE #38: Add Watchdog/Self-Health Supervisor
The Zephyr agent ([Aeolus](https://github.com/absmach/aeolus)) has a hardware watchdog timer ([`src/management/reset.c`](https://github.com/absmach/aeolus/blob/main/src/management/reset.c)) that reboots the device if the agent thread or system workqueue hangs. Two feeders (work item + main loop) ensure either thread stall is detected.

On Linux, the equivalent would be a self-health-check goroutine or integration with the systemd watchdog.

**What to implement:**

- Option A: Systemd watchdog integration — `WatchdogSec=` in the unit file, periodic `sd_notify("WATCHDOG=1")` from a goroutine
- Option B: Internal health supervisor goroutine that checks MQTT connection, HTTP server, and subsystem liveness; restarts the process if unhealthy for too long
- Option A is recommended when running under systemd; Option B for containerized deployments

**Reference:** Aeolus watchdog — [`src/management/reset.c`](https://github.com/absmach/aeolus/blob/main/src/management/reset.c), [`docs/watchdog.md`](https://github.com/absmach/aeolus/blob/main/docs/watchdog.md)

rodneyosodo:
## What does this do?

Adds a health watchdog/self-health supervisor that monitors subsystem liveness and restarts the agent when unhealthy for too long.

The supervisor is a background goroutine that periodically checks registered health checkers (currently MQTT connection status). If any checker reports unhealthy for longer than the configured timeout, the agent re-executes itself via `syscall.Exec`.

When running under systemd with `Type=notify`, the supervisor also sends periodic `WATCHDOG=1` notifications through the `NOTIFY_SOCKET` unix datagram socket.

The feature is disabled by default (`MG_AGENT_WATCHDOG_INTERVAL=0s`). Enable with:
```bash
MG_AGENT_WATCHDOG_INTERVAL=10s   # check every 10s MG_AGENT_WATCHDOG_TIMEOUT=60s    # restart after 60s unhealthy
```

## Which issue(s) does this PR fix/relate to?

Resolves #38

## List any changes that modify/break current functionality

None. The watchdog is disabled by default. When enabled, it only affects behavior if the agent becomes permanently unhealthy — previously the process would hang silently; now it restarts.

## Have you included tests for your changes?

Yes — `pkg/health/health_test.go` covers: disabled supervisor, healthy state, unhealthy detection, and the `MQTTChecker`.

## Did you document any new/modified functionality?

The new env vars (`MG_AGENT_WATCHDOG_INTERVAL`, `MG_AGENT_WATCHDOG_TIMEOUT`) are self-documenting. The supervisor logs health state transitions and systemd watchdog integration.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -225,10 +225,15 @@ func main() {
 	// the process if the agent remains unhealthy for longer than the timeout.
 	// When running under systemd, it also sends periodic WATCHDOG=1
 	// notifications via the NOTIFY_SOCKET.
-	wdInterval, _ := time.ParseDuration(c.WatchdogInterval)
-	wdTimeout, _ := time.ParseDuration(c.WatchdogTimeout)
-	if wdInterval <= 0 {
-		wdTimeout, _ = time.ParseDuration("60s")
+	wdInterval, err := time.ParseDuration(c.WatchdogInterval)
+	if err != nil {
+		logger.Warn("Invalid watchdog interval, disabling watchdog", slog.String("interval", c.WatchdogInterval), slog.Any("error", err))
+		wdInterval = 0
+	}
+	wdTimeout, err := time.ParseDuration(c.WatchdogTimeout)
+	if err != nil {
+		logger.Warn("Invalid watchdog timeout, using default 60s", slog.String("timeout", c.WatchdogTimeout), slog.Any("error", err))
+		wdTimeout = 60 * time.Second
 	}
 	sup := health.NewSupervisor(health.Config{
 		Interval: wdInterval,
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 	pkgconfig "github.com/absmach/agent/pkg/config"
 	"github.com/absmach/agent/pkg/conn"
 	"github.com/absmach/agent/pkg/devicemgr"
+	"github.com/absmach/agent/pkg/health"
 	"github.com/absmach/agent/pkg/iface"
 	"github.com/absmach/agent/pkg/logstream"
 	"github.com/absmach/agent/pkg/nodered"
@@ -69,6 +70,8 @@ type config struct {
 	DeviceDBPath         string `env:"MG_AGENT_DEVICE_DB_PATH"                envDefault:"/var/lib/agent/devices.db"`
 	ConfigPath           string `env:"MG_AGENT_CONFIG_PATH"                   envDefault:"agent-config.json"`
 	CommandSecret        string `env:"MG_AGENT_COMMAND_SECRET"                envDefault:""`
+	WatchdogInterval     string `env:"MG_AGENT_WATCHDOG_INTERVAL"            envDefault:"0s"`
+	WatchdogTimeout      string `env:"MG_AGENT_WATCHDOG_TIMEOUT"             envDefault:"60s"`
 }
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -218,6 +221,21 @@ func main() {
...
 	onReconnect = b.Resubscribe
 
+	// Health supervisor: periodically checks subsystem liveness and restarts
+	// the process if the agent remains unhealthy for longer than the timeout.
+	// When running under systemd, it also sends periodic WATCHDOG=1
+	// notifications via the NOTIFY_SOCKET.
+	wdInterval, _ := time.ParseDuration(c.WatchdogInterval)
+	wdTimeout, _ := time.ParseDuration(c.WatchdogTimeout)
+	if wdInterval <= 0 {
+		wdTimeout, _ = time.ParseDuration("60s")
```
JeffMboya: Resetting `wdTimeout` here is dead code; this branch only runs when the supervisor is disabled.

cmd/main.go
```diff
@@ -218,6 +221,21 @@ func main() {
 	b := conn.NewBroker(svc, mqttClient, cfg.Channels.CtrlChan(), cfg.DomainID, logger)
 	onReconnect = b.Resubscribe
 
+	// Health supervisor: periodically checks subsystem liveness and restarts
+	// the process if the agent remains unhealthy for longer than the timeout.
+	// When running under systemd, it also sends periodic WATCHDOG=1
+	// notifications via the NOTIFY_SOCKET.
+	wdInterval, _ := time.ParseDuration(c.WatchdogInterval)
```
JeffMboya: Parse errors silently discarded; an invalid env value silently disables the watchdog.

pkg/health/health.go
```diff
@@ -0,0 +1,185 @@
...
+		case <-ticker.C:
+			anyUnhealthy := false
+			for _, c := range s.checkers {
+				if !c.Healthy() {
+					s.logger.Warn("Health check failed", slog.String("checker", c.Name()))
+					anyUnhealthy = true
+				}
+			}
+
+			if anyUnhealthy {
```
JeffMboya, line 103: `s.healthy` is never set `false` here; `IsHealthy()` always returns `true` even when unhealthy.

pkg/health/health.go
```diff
@@ -0,0 +1,185 @@
...
+func (s *Supervisor) sdWatchdogLoop(ctx context.Context, socketPath string) {
+	ticker := time.NewTicker(s.interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			if err := sdNotify(socketPath, "WATCHDOG=1\n"); err != nil {
```
JeffMboya: `WATCHDOG=1` sent unconditionally; systemd should only be notified when the agent is actually healthy.

pkg/health/health.go
```diff
@@ -0,0 +1,185 @@
...
+				s.logger.Warn("Failed to notify systemd watchdog", slog.Any("error", err))
+			}
+		}
+	}
+}
+
+// sdNotify sends a message to the systemd notification socket.
+func sdNotify(socketPath, state string) error {
+	addr := &net.UnixAddr{Name: socketPath, Net: "unixgram"}
+	conn, err := net.DialUnix("unixgram", nil, addr)
```
JeffMboya: New UDP connection per tick is wasteful; reuse a single persistent connection instead.

JeffMboya (approved):
LGTM

PR #91, opened 2026-06-13
AG-81 - Enhance Command Registry with Extensible Dispatch System

### LINKED ISSUE #81: Enhance Command Registry with Extensible Dispatch System
## Overview

Extend the command dispatch system to support a wider range of commands similar to Aeolus, including exec, config, control, reset, ota, ping, route, and term.

## Description

Currently the Linux agent supports: `control`, `exec`, `config`, `term`, `nodered`

Aeolus supports additional commands:
- `exec` - Execute gateway actions (GPIO, LED, etc.)
- `config` - View or save runtime config
- `control` - Internal agent control
- `reset` - Reboot device
- `ota` - Trigger or query OTA update
- `ping` - Health check response
- `route` - Forward command to downstream device
- `term` - Terminal session

## Requirements

1. **Command Registry**
   - Extensible command registration system
   - Plugin-style architecture
   - Command metadata (name, description, args)
   - Command help/usage documentation

2. **Command Implementations**
   - `exec` - Execute shell commands (already exists, enhance)
   - `config` - View/save config (already exists, enhance)
   - `control` - Agent lifecycle control (start/stop/reload)
   - `reset` - Graceful or immediate reboot
   - `ota` - OTA management (query/status/confirm)
   - `ping` - Health check with diagnostics
   - `route` - Route to downstream devices (see #7)
   - `term` - Terminal session (already exists, enhance)

3. **Command Parsing**
   - SenML argument parsing
   - Argument validation
   - Error handling
   - Response formatting

4. **Authorization**
   - Per-command authorization
   - Token-based authentication
   - Audit logging

## Command Format

```json
[
  {
    "bn": "cmd-42:",
    "n": "exec",
    "vs": "gpio_set,led0,1"
  }
]
```

## Reference

See Aeolus implementation:
- Command dispatch table in `docs/agent.md`
- `src/control/`
- `src/agent.c` command handler

## Acceptance Criteria

- [ ] Command registry can register new commands dynamically
- [ ] All 8 commands are implemented
- [ ] Commands validate arguments
- [ ] Commands return proper SenML responses
- [ ] `control` command can start/stop/reload agent
- [ ] `reset` command supports graceful and immediate modes
- [ ] `ota` command can query status and trigger updates
- [ ] `ping` command returns health diagnostics
- [ ] Authorization is enforced per command
- [ ] Integration tests for each command

## Priority

High

JeffMboya:
### What does this do?
Adds an extensible command registry with metadata, a help command, per-command auth, plus route, control lifecycle, and ota status commands.

Resolves #81

None breaking. Authorization moves from a global gate to per-command; all built-ins still require a token.

Yes — registry, per-command auth, help, ota status, route validation, and all control subcommands are covered.

Yes — README command table and usage examples updated for help, control, ota status, and route.

### Notes
New commands are MQTT-only (SenML), matching the issue format; no HTTP endpoints added.

### DIFF
--- a/service.go
+++ b/service.go
@@ -53,6 +53,12 @@ const (
 	control = "control"
 	data    = "data"
 
+	// control subcommands for agent lifecycle management.
+	ctrlStop   = "stop"
+	ctrlStart  = "start"
+	ctrlReload = "reload"
+	ctrlStatus = "status"
+
 	export = "export"
 
 	keyLogLevel               = "log_level"
@@ -140,6 +146,9 @@ var (
 
 	// errDeviceManagerDisabled indicates device manager is not configured.
 	errDeviceManagerDisabled = errors.New("device manager not configured")
+
+	// errRouteFailed indicates a failure routing a command to a downstream device.
+	errRouteFailed = errors.New("failed to route command to device")
 )
 
 // DeviceService groups the downstream-device management methods of Service.
@@ -171,6 +180,11 @@ type Service interface {
 
 	// Control command.
 	Control(uuid, cmdStr string) error
+
+	// Route forwards a raw payload to a downstream device's physical interface
+	// and publishes the device's response. cmdStr is comma-delimited:
+	// <device_id>,<hex_payload>[,<read_bytes>].
+	Route(ctx context.Context, uuid, cmdStr string) error
 
 	// Update configuration file.
 	AddConfig(Config) error
@@ -264,6 +278,8 @@ type agent struct {
 	otaCancel           context.CancelFunc
 	otaAborted          atomic.Bool
 	bootstrapCachePath  string
+	runCtx              context.Context
+	paused              atomic.Bool
 }
 
 // New returns agent service implementation.
@@ -284,6 +300,7 @@ func New(ctx context.Context, mc paho.Client, cfg *Config, nc nodered.Client, lo
 		logLevel:            levelVar,
 		startupConfig:       *cfg,
 		bootstrapCachePath:  bootstrapCachePath,
+		runCtx:              ctx,
 	}
 
 	if devices != nil {
@@ -377,6 +394,14 @@ func (a *agent) Control(uuid, cmdStr string) error {
 
 	cmd := cmdArgs[0]
 	switch {
+	case cmd == ctrlStop:
+		resp = a.controlStop()
+	case cmd == ctrlStart:
+		resp = a.controlStart()
+	case cmd == ctrlReload:
+		resp = a.controlReload()
+	case cmd == ctrlStatus:
+		resp, err = a.controlStatus()
 	case strings.HasPrefix(cmd, "nodered-"):
 		resp, err = a.NodeRed(cmdStr)
 	default:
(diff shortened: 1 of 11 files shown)


rodneyosodo (commented):
Looks good, just a few minor things

service.go
```diff
@@ -384,6 +409,120 @@ func (a *agent) Control(uuid, cmdStr string) error {
...
+func (a *agent) Route(_ context.Context, uuid, cmdStr string) error {
+	if a.devices == nil {
+		return errors.Wrap(errRouteFailed, errDeviceManagerDisabled)
+	}
+
+	args := strings.Split(strings.ReplaceAll(cmdStr, " ", ""), ",")
+	if len(args) < 2 || args[0] == "" || args[1] == "" {
+		return errors.Wrap(errRouteFailed, errInvalidCommand)
+	}
+	deviceID, hexPayload := args[0], args[1]
```
rodneyosodo, line 531: **Issue 1: Error Handling in Route**

The Route command publishes errors to the "control" topic but doesn't validate device existence before attempting operations. This could lead to confusing error messages when a device doesn't exist vs when there's an actual interface error.

**Recommendation:** Consider adding device existence validation before opening interfaces to provide clearer error messages for operators to diagnose issues.

service.go
```diff
@@ -384,6 +409,120 @@ func (a *agent) Control(uuid, cmdStr string) error {
...
+	return "started"
+}
+
+// controlReload re-applies persisted runtime config overrides from the store so
+// that out-of-band changes to the store take effect without a restart.
+func (a *agent) controlReload() string {
+	if a.store == nil {
+		return notConfigured
+	}
+	for key, val := range a.store.All() {
```
rodneyosodo, line 470: **Issue 2: Configuration Reload Validation**

The `controlReload` function applies persisted values without validating they match expected types/ranges. If invalid values were persisted (e.g., through manual file editing), this could cause runtime issues.

**Recommendation:** Add validation before applying values and return which keys were updated for auditability.

service.go
```diff
@@ -384,6 +409,120 @@ func (a *agent) Control(uuid, cmdStr string) error {
...
+	}
+	a.logger.Info("Agent paused via control command")
+	return "stopped"
+}
+
+// controlStart resumes the background publishing loops and restarts the
+// per-device scheduler using the agent's run context.
+func (a *agent) controlStart() string {
+	a.paused.Store(false)
+	if a.sched != nil && a.runCtx != nil {
```
rodneyosodo: **Issue 3: Nil runCtx Risk in controlStart**

The `runCtx` field is set to the `ctx` parameter in `New()`, but if this context is canceled before `controlStart` is called, restarting the scheduler will fail. This could leave the agent in an inconsistent state where stop works but start fails.

**Recommendation:** Add a nil check and consider using a separate lifecycle context that isn't tied to the initialization context.

service.go
JeffMboya: Done in f5bf596 — `Route` now resolves the device with `devices.Get(deviceID)` up front, so a missing device returns a clear not-found error before any interface open is attempted.

service.go
JeffMboya: Done in f5bf596 — each persisted value is now validated via the existing `validateSettableValue` before being applied; invalid entries are skipped and logged, and the response lists the applied keys for auditability (e.g. `reloaded:log_level`).

service.go
JeffMboya: Done in f5bf596 — `controlStart` now skips the scheduler restart when `runCtx` is nil or already cancelled and logs the reason instead of failing silently. A fully independent lifecycle context would leak device goroutines past process shutdown (the scheduler ctx must descend from the process ctx), so the explicit cancelled-context guard is the safer fix here.

rodneyosodo (approved):

PR #92, opened 2026-06-17
AGT-79 - Publish OTA Progress and Status to MQTT

### LINKED ISSUE #79: Publish OTA Progress and Status to MQTT
## Overview

Implement OTA status topic publishing to report download progress, verification status, and update state over MQTT.

## Description

Aeolus publishes OTA progress to the `ota/status` topic so administrators can monitor update progress in real-time. The current Linux OTA implementation exists but doesn't publish status to MQTT.

## Requirements

1. **Status Topic**
   - Publish to `m/{domain}/c/{ctrl_channel}/ota/status`
   - Update on state transitions
   - Update on download progress (every 5%)
   - Use SenML format

2. **Status Fields**
   - `state` - Current state (idle, downloading, verifying, ready, restarting, aborted)
   - `bytes` - Bytes downloaded so far
   - `total` - Total bytes expected
   - `progress` - Percentage complete (0-100)
   - `error` - Error message (if any)

3. **State Publishing**
   - Publish on each state transition
   - Publish periodically during download
   - Publish final status on completion/abort
   - Retain last status message

4. **OTA Config Topic**
   - Subscribe to `m/{domain}/c/{ctrl_channel}/ota/cfg`
   - Parse update trigger
   - Support URL, hash, and size fields

5. **OTA Data Topic**
   - Subscribe to `m/{domain}/c/{ctrl_channel}/ota`
   - Receive firmware binary via MQTT streaming
   - Alternative to HTTP download

## Status Format

```json
[
  {
    "n": "state",
    "vs": "downloading"
  },
  {
    "n": "bytes",
    "v": 65536,
    "u": "By"
  },
  {
    "n": "total",
    "v": 1324740,
    "u": "By"
  }
]
```

## Reference

See Aeolus implementation:
- `src/management/ota.c`
- `docs/ota.md`
- Current OTA implementation in `pkg/ota/ota.go`

## Acceptance Criteria

- [ ] OTA status topic publishes state changes
- [ ] Download progress publishes every 5%
- [ ] Status includes bytes downloaded and total
- [ ] Final status published on completion
- [ ] Error status published on failure
- [ ] OTA can be triggered via MQTT config topic
- [ ] MQTT streaming download works
- [ ] HTTP download still works
- [ ] ProgressFn callback integrated with MQTT publisher
- [ ] Integration tests for status publishing

## Priority
High

nyagamunene:
### What does this do?

Implements OTA status publishing to MQTT as described in #79. The agent now reports download progress, verification status, and update state in real-time over MQTT, and supports receiving firmware binaries directly via MQTT as an alternative to HTTP download.

**Status topic publishing (`m/{domain}/c/{ctrl_channel}/ota/status`)**
- Publishes on every state transition: `triggered → downloading → verifying → ready → restarting` (or `aborted` on failure)
- Reports download progress every 5% with `bytes`, `total`, and `progress` fields
- Uses `retain=true` so the broker always holds the last known OTA state
- Includes an `error` field in the final status message on failure or abort
- SenML fields match the spec: `state` (string), `bytes` (By), `total` (By), `progress` (%)

**MQTT binary delivery (`m/{domain}/c/{ctrl_channel}/ota`)**
- New `ota/cfg` flow: when `url` is absent but `hash` and `size` are present, the broker primes a pending data receiver
- Firmware binary arrives as a raw MQTT payload on the `ota` topic
- Agent verifies the SHA-256 hash and installs — same verify/replace path as HTTP download
- HTTP download path is unchanged

**API / internals**
- `ProgressFn` extended to include `bytesWritten, totalBytes int64` alongside `progress float64`
- New `ParseCfgFromRecords()` in `pkg/ota` — parses cfg records without requiring a URL
- New `RunFromData()` in `pkg/ota` — installs firmware from an in-memory byte slice
- New `OTAFromData(ctx, data, sha256hex)` method on `Service` interface (+ logging/metrics middleware)

Resolves #79

- `ota.ProgressFn` signature changed: `func(State, float64)` → `func(State, int64, int64, float64)`. Any external code calling `ota.Run()` with a progress callback must update the callback signature.
- OTA status SenML field names changed: `ota_state` → `state`, `ota_progress` → `progress`. New fields `bytes` and `total` are added.
- OTA status messages now use `retain=true` (previously `false`).

Yes:
- Updated all existing `ota_test.go` tests for the new `ProgressFn` signature
- `TestParseCfgFromRecords` — covers the new URL-optional cfg parser
- `TestRunFromData` — covers MQTT binary delivery path (empty hash, hash mismatch, valid hash)
- `TestOTAStatusPublishing` — verifies correct SenML fields on the status topic
- `TestOTAStatusPublishesErrorOnFailure` — verifies `error` field is included in the final failure status
- `TestOTAFromDataStatusPublishing` — verifies status publishing for the MQTT data-delivery path

The `ota/cfg` and `ota` topic behaviour is described in the inline code comments and follows the protocol defined in #79.

### Notes

`syscall.Exec` is called on a successful OTA install (process is replaced in-place). Tests that exercise the full verify+replace path intentionally use invalid binary content so `syscall.Exec` returns `ENOEXEC` rather than replacing the test process.

### DIFF
--- a/pkg/ota/ota.go
+++ b/pkg/ota/ota.go
@@ -6,7 +6,6 @@ package ota
 import (
 	"context"
 	"crypto/sha256"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -61,8 +60,9 @@ func (s State) String() string {
 }
 
 // ProgressFn is called on state transitions and during download.
-// bytesWritten and totalBytes are bytes on disk; they are only meaningful during
-// StateDownloading. progress is 0–100 and is only meaningful in StateDownloading.
+// bytesWritten and totalBytes are bytes on disk; during StateDownloading they
+// reflect the running count, and in all later states they report the final file
+// size. progress is 0–100 and is only meaningful in StateDownloading.
 type ProgressFn func(state State, bytesWritten, totalBytes int64, progress float64)
 
 // Config holds the OTA updater parameters.
@@ -98,7 +98,7 @@ func TriggerFromRecords(records []senml.Record) (Trigger, error) {
 			t.URL = strings.TrimSpace(*r.StringValue)
 		case fieldHash:
 			if r.StringValue != nil {
-				t.SHA256Hex = *r.StringValue
+				t.SHA256Hex = strings.TrimSpace(*r.StringValue)
 			}
 		case fieldSize:
 			if r.Value != nil {
@@ -125,7 +125,7 @@ func ParseCfgFromRecords(records []senml.Record) Trigger {
 			}
 		case fieldHash:
 			if r.StringValue != nil {
-				t.SHA256Hex = *r.StringValue
+				t.SHA256Hex = strings.TrimSpace(*r.StringValue)
 			}
 		case fieldSize:
 			if r.Value != nil {
@@ -153,19 +153,20 @@ func Run(ctx context.Context, cfg Config, url, sha256hex string, size uint64, pr
 		progressFn(StateDownloading, written, total, pct)
 	})
 	if err != nil {
-		if errors.Is(err, context.Canceled) {
-			progressFn(StateAborted, 0, 0, 0)
-		}
 		return fmt.Errorf("ota download: %w", err)
 	}
 
-	progressFn(StateVerifying, 0, 0, 100)
+	// Report the actual file size for subsequent states so that subscribers
+	// see consistent bytes/total values across HTTP- and MQTT-delivered updates.
+	var totalBytes int64
+	if fi, statErr := os.Stat(tmpPath); statErr == nil {
+		totalBytes = fi.Size()
+	}
+
+	progressFn(StateVerifying, totalBytes, totalBytes, 100)
 	verified, err := verify(ctx, url, tmpPath, sha256hex)
 	if err != nil {
 		_ = os.Remove(tmpPath)
-		if errors.Is(err, context.Canceled) {
-			progressFn(StateAborted, 0, 0, 0)
-		}
 		return fmt.Errorf("ota verify: %w", err)
 	}
 	if !verified {
@@ -173,13 +174,13 @@ func Run(ctx context.Context, cfg Config, url, sha256hex string, size uint64, pr
 		return fmt.Errorf("ota verify: no hash provided and sidecar not found at %s.sha256; refusing unverified install", url)
 	}
 
-	progressFn(StateReady, 0, 0, 100)
+	progressFn(StateReady, totalBytes, totalBytes, 100)
 	if err := replace(tmpPath, cfg.BinaryPath); err != nil {
 		_ = os.Remove(tmpPath)
 		return fmt.Errorf("ota replace: %w", err)
 	}
 
-	progressFn(StateRestarting, 0, 0, 100)
+	progressFn(StateRestarting, totalBytes, totalBytes, 100)
 	return syscall.Exec(cfg.BinaryPath, os.Args, os.Environ())
 }
 
@@ -198,9 +199,12 @@ func RunFromData(ctx context.Context, cfg Config, data []byte, sha256hex string,
 	progressFn(StateTriggered, 0, totalBytes, 0)
 	progressFn(StateDownloading, 0, totalBytes, 0)
 
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("ota write: %w", err)
+	}
+
 	f, err := os.CreateTemp(cfg.DownloadDir, "agent-ota-*")
 	if err != nil {
-		progressFn(StateAborted, 0, totalBytes, 0)
 		return fmt.Errorf("ota create temp: %w", err)
 	}
 	tmpPath := f.Name()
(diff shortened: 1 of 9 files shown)


pkg/ota/ota.go
```diff
@@ -148,20 +173,72 @@ func Run(ctx context.Context, cfg Config, url, sha256hex string, size uint64, pr
...
+	}
+
+	progressFn(StateRestarting, 0, 0, 100)
+	return syscall.Exec(cfg.BinaryPath, os.Args, os.Environ())
+}
+
+// RunFromData installs firmware from an in-memory binary payload instead of
+// downloading from HTTP. sha256hex must be the hex-encoded SHA-256 of data.
+// On success it never returns (the process is replaced in-place).
+func RunFromData(ctx context.Context, cfg Config, data []byte, sha256hex string, progressFn ProgressFn) error {
```
rodneyosodo, line 190: **`RunFromData` ignores `ctx` — context cancellation not respected during file I/O and hash verification**

`RunFromData` accepts `ctx context.Context` but never reads from it. Unlike `Run`, which passes `ctx` into `download` (where `ctx.Err()` is checked on every read loop at line 279), `RunFromData` performs file I/O and SHA-256 verification without any context check. If the context is cancelled (e.g., user sends `ota abort` during an MQTT-delivered update), the write-to-disk and hash operations run to completion before the function returns.

Consider checking `ctx.Err()` at least before/after the file I/O section (lines 208-217) and inside `verifyFile` (or wrap the file read with a context-aware reader).

pkg/ota/ota.go
```diff
@@ -111,6 +112,30 @@ func TriggerFromRecords(records []senml.Record) (Trigger, error) {
...
+	var t Trigger
+	for _, r := range records {
+		switch r.Name {
+		case fieldURL:
+			if r.StringValue != nil {
+				t.URL = strings.TrimSpace(*r.StringValue)
+			}
+		case fieldHash:
+			if r.StringValue != nil {
+				t.SHA256Hex = *r.StringValue
```
rodneyosodo: **`ParseCfgFromRecords` doesn't trim hash field**

`ParseCfgFromRecords` trims `url` via `strings.TrimSpace` (line 124) but reads `hash` as-is on line 128. `TriggerFromRecords` at line 101 has the same omission, so this is consistent with existing code. Low severity since whitespace in a hash won't match — but worth noting for consistency.

pkg/ota/ota.go
```diff
@@ -119,27 +144,27 @@ func TriggerFromRecords(records []senml.Record) (Trigger, error) {
...
 	if err != nil {
 		if errors.Is(err, context.Canceled) {
-			progressFn(StateAborted, 0)
+			progressFn(StateAborted, 0, 0, 0)
 		}
 		return fmt.Errorf("ota download: %w", err)
 	}
 
-	progressFn(StateVerifying, 100)
+	progressFn(StateVerifying, 0, 0, 100)
```
rodneyosodo: **Inconsistent `bytes`/`total` between `Run` and `RunFromData`**

In `Run`, the `StateVerifying` (line 162), `StateReady` (line 176), and `StateRestarting` (line 182) callbacks pass `0, 0` for `bytesWritten`/`totalBytes`. In `RunFromData`, the same states pass `totalBytes, totalBytes` (lines 220, 227, 233).

A subscriber reading the status topic after an MQTT-delivered update sees `"bytes": 1324740` during verifying while an HTTP-delivered update sees `"bytes": 0`. This is by design per the `ProgressFn` doc comment ("meaningful during downloading"), but the inconsistency could confuse consumers that inspect the `bytes` field. Consider having `Run` also report the actual file size after download (a `os.Stat` call after download completes would give it).

rodneyosodo (commented):
The core logic (MQTT data priming, SHA-256 verification, status publishing) is sound. Tests cover the key paths. No security issues or breaking behavioural regressions.

service.go
```diff
@@ -1343,16 +1341,101 @@ func (a *agent) OTA(ctx context.Context, url, sha256hex string, size uint64) err
...
+
+	publishStatus := func(state ota.State, bytesWritten, totalBytes int64, progress float64, errMsg string) {
+		a.publishOTAStatus(statusTopic, qos, state, bytesWritten, totalBytes, progress, errMsg)
+	}
+	progressFn := func(state ota.State, bytesWritten, totalBytes int64, progress float64) {
+		publishStatus(state, bytesWritten, totalBytes, progress, "")
+	}
+
+	runErr := ota.RunFromData(otaCtx, otaCfg, data, sha256hex, progressFn)
+	if runErr != nil {
```
rodneyosodo, line 1400: **Double `StateAborted` publish**

Both the OTA library functions (`Run` at lines 157, 167; `RunFromData` at lines 203, 211, 223) and the service layer (this function at lines 1400-1409) independently publish `StateAborted`. This means two consecutive MQTT messages are published for a single failure:

1. From `RunFromData`: `progressFn(StateAborted, …)` → MQTT message **without** `error` field
2. From `service.go` line 1403/1406: `publishStatus(StateAborted, …, errMsg)` → MQTT message **with** `error` field

Since the messages use `retain=true`, the second (correct) message replaces the first. But subscribers consuming in real-time will see a transient "aborted" without an error before the final one arrives. Consider removing the internal `progressFn(StateAborted, …)` calls from `Run` and `RunFromData` and letting the service layer be the sole publisher of the aborted state.

rodneyosodo (approved):

PR #98, opened 2026-06-23
AG-73 - Add device registry backup/restore, migrations, and lifecycle webhooks

### LINKED ISSUE #73: Implement Device Manager/Registry for Downstream Devices
## Overview

Implement a dynamic device registry similar to Aeolus that can manage 100+ downstream devices with persistent storage.

## Description

The Aeolus Zephyr agent includes a comprehensive device manager that:
- Maintains a registry of downstream devices (BLE, Zigbee/Thread, I2C, UART/Modbus)
- Supports dynamic provisioning via API
- Routes commands to the correct interface
- Persists device registry to NVS

## Requirements

1. **Device Registry**
   - Store device metadata (ID, type, interface, config)
   - Support 100+ concurrent devices
   - Persistent storage (file-based for Linux)
   - CRUD operations via HTTP API

2. **Dynamic Provisioning**
   - REST API for adding/removing devices
   - Webhook support for device lifecycle events
   - Device authentication/authorization

3. **Interface Routing**
   - Route commands to specific device interfaces
   - Handle device responses
   - Support for multiple protocol adapters

4. **Storage**
   - Persistent JSON file or SQLite database
   - Backup/restore functionality
   - Migration support

## Reference

See Aeolus implementation:
- `src/device/device_manager.c`
- `src/device/device_registry.c`
- `docs/provisioning.md`

## Acceptance Criteria

- [ ] Device can be added via API
- [ ] Device can be queried by ID
- [ ] Device can be removed
- [ ] Registry persists across restarts
- [ ] Commands can be routed to registered devices
- [ ] API documentation included

## Priority

High

JeffMboya:
### What does this do?

Completes the remaining requirements of the downstream device manager (issue #73) that were not yet present in `main`. The core registry — persistent BoltDB storage, full CRUD over HTTP + MQTT, interface routing, and Magistrala provisioning — already exists; this PR adds the three outstanding items from the issue's Requirements:

- **Schema migrations** — the device store now tracks an on-disk `schema_version` in a `meta` bucket and runs pending migrations forward inside a single transaction on open. Databases from older agent builds (no version) are treated as v0 and migrated in place without data loss; a database from a newer schema than the running binary is rejected.
- **Backup/restore** — the registry can be exported to a portable JSON snapshot (schema version, timestamp, devices) and re-imported, either merging (overwrite by ID) or replacing the registry. Exposed via `GET /devices/backup` and `POST /devices/restore?replace=true`.
- **Lifecycle webhooks** — when `MG_AGENT_DEVICE_WEBHOOK_URL` is set, the agent asynchronously POSTs `device.added` / `device.removed` / `device.seen` / `device.iface_opened` / `device.iface_closed` events to that endpoint. Delivery is best-effort from a background worker so it never blocks device operations; an optional `MG_AGENT_DEVICE_WEBHOOK_SECRET` is sent as `X-Agent-Webhook-Secret`.

Resolves #73

None. All additions are backward compatible:
- `devicemgr.New` gains a variadic `...Option` argument, so existing callers compile unchanged; webhooks are wired through `WithWebhook` and default to a no-op notifier.
- The `Service` interface gains `BackupDevices` and `RestoreDevices`; logging/metrics middleware and the generated mock are updated accordingly.
- Existing device databases are migrated forward automatically on first open.

Yes:
- `pkg/devicemgr/migration_test.go` — schema version + migration persistence across reopen.
- `pkg/devicemgr/backup_test.go` — store export/import (merge, replace, validation) and manager backup/restore.
- `pkg/devicemgr/webhook_test.go` — lifecycle event emission, secret header, no-op default, empty-URL disable.
- `api/endpoints_test.go` — `GET /devices/backup` and `POST /devices/restore` handlers.

`go build ./...`, `go vet ./...`, `gofmt -l`, `go test ./...`, and `go test -race ./pkg/devicemgr/` all pass.

Yes — `docs/devices.md` documents persistence/migrations, backup/restore (with `curl` examples), lifecycle webhooks (event table + payload), the new environment variables, and the expanded HTTP API table.

### DIFF
--- a/api/requests.go
+++ b/api/requests.go
@@ -3,7 +3,10 @@
 
 package api
 
-import "github.com/absmach/agent"
+import (
+	"github.com/absmach/agent"
+	"github.com/absmach/agent/pkg/devicemgr"
+)
 
 type serverConfig struct {
 	Port string `json:"port"`
@@ -117,6 +120,20 @@ type addDeviceReq struct {
 func (req addDeviceReq) validate() error {
 	if req.Name == "" || req.ExtID == "" || req.ExtKey == "" || req.IfaceType == "" || req.IfaceAddr == "" {
 		return agent.ErrMalformedEntity
+	}
+	return nil
+}
+
+type restoreDevicesReq struct {
+	Backup  devicemgr.Backup
+	Replace bool
+}
+
+func (req restoreDevicesReq) validate() error {
+	for _, d := range req.Backup.Devices {
+		if d.ID == "" {
+			return agent.ErrMalformedEntity
+		}
 	}
 	return nil
 }
--- a/api/transport.go
+++ b/api/transport.go
@@ -107,6 +107,18 @@ func MakeHandler(svc agent.Service, logger *slog.Logger, stream *logstream.Strea
 		EncodeResponse,
 		opts...,
 	).ServeHTTP)
+	r.Get("/devices/backup", kithttp.NewServer(
+		backupDevicesEndpoint(svc),
+		decodeRequest,
+		EncodeResponse,
+		opts...,
+	).ServeHTTP)
+	r.Post("/devices/restore", kithttp.NewServer(
+		restoreDevicesEndpoint(svc),
+		decodeRestoreDevicesRequest,
+		EncodeResponse,
+		opts...,
+	).ServeHTTP)
 	r.Get("/devices/{id}", kithttp.NewServer(
 		getDeviceEndpoint(svc),
 		decodeIDFromPath,
@@ -260,6 +272,21 @@ func decodeNodeRedRequest(_ context.Context, r *http.Request) (any, error) {
 func decodeAddDeviceRequest(_ context.Context, r *http.Request) (any, error) {
 	req := addDeviceReq{}
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return nil, mgerrors.Wrap(apiutil.ErrMalformedRequestBody, err)
+	}
+	return req, nil
+}
+
+// maxRestoreBodyBytes caps the device-registry restore upload (16 MiB is ample
+// for tens of thousands of device records) so a malformed or hostile request
+// cannot exhaust memory.
+const maxRestoreBodyBytes = 16 << 20
+
+func decodeRestoreDevicesRequest(_ context.Context, r *http.Request) (any, error) {
+	req := restoreDevicesReq{
+		Replace: strings.EqualFold(r.URL.Query().Get("replace"), "true"),
+	}
+	if err := json.NewDecoder(io.LimitReader(r.Body, maxRestoreBodyBytes)).Decode(&req.Backup); err != nil {
 		return nil, mgerrors.Wrap(apiutil.ErrMalformedRequestBody, err)
 	}
 	return req, nil
--- a/docs/devices.md
+++ b/docs/devices.md
@@ -57,16 +57,92 @@ When a device has a valid channel ID, the agent launches a background goroutine 
 
 Reconnection uses exponential backoff (1s to 30s). TLS settings are inherited from the gateway's MQTT configuration.
 
+## Persistence and Migrations
+
+The registry is stored in a BoltDB file (`MG_AGENT_DEVICE_DB_PATH`) so it survives agent restarts. The store tracks an on-disk **schema version** in a `meta` bucket. When the agent opens the database it runs any pending migrations forward to the current schema version inside a single write transaction, so upgrading the agent never loses devices. Databases created by older agent builds (which had no schema version) are treated as version `0` and migrated in place on first open. A database written by a newer agent than the running binary is rejected rather than silently downgraded.
+
+## Backup and Restore
+
+The entire registry can be exported to a portable JSON snapshot and re-imported, either into the same agent or a different one. The snapshot carries the schema version and an export timestamp:
+
+```json
+{
+  "schema_version": 1,
+  "exported_at": "2026-06-23T08:00:00Z",
+  "devices": [{ "id": "...", "name": "...", "interface_type": "serial", "...": "..." }]
+}
+```
+
+- **Backup** — `GET /devices/backup` returns the snapshot.
+- **Restore** — `POST /devices/restore` imports a snapshot. By default devices are **merged** (records with the same ID are overwritten, others are left untouched). Pass `?replace=true` to first clear the existing registry so it matches the snapshot exactly; any open physical interfaces are closed during a replace restore. A snapshot from a newer schema version is rejected, and devices with an empty ID are rejected with `400`. The upload body is capped at 16 MiB.
+
+> A replace restore is an administrative operation and is not synchronized against concurrent interface opens. If a device interface is opened at the same moment a replace restore wipes that device, the interface can be left open with no owning record. Quiesce device activity (or restart the agent) around a replace restore.
+
+```bash
+# Back up the registry to a file
+curl -s http://localhost:9999/devices/backup | jq . > devices-backup.json
+
+# Restore (merge) on another agent
+curl -s -X POST http://localhost:9999/devices/restore \
+  -H 'Content-Type: application/json' \
+  --data-binary @devices-backup.json
+
+# Restore and replace the existing registry
+curl -s -X POST 'http://localhost:9999/devices/restore?replace=true' \
+  -H 'Content-Type: application/json' \
+  --data-binary @devices-backup.json
+```
+
+## Lifecycle Webhooks
+
+When `MG_AGENT_DEVICE_WEBHOOK_URL` is configured, the agent POSTs a JSON event to that URL on device lifecycle changes. Delivery is asynchronous and best-effort: events are queued and sent from a background worker, so webhook latency never blocks device operations. A failed delivery (network error or `5xx`) is retried twice with a short backoff; if the queue overflows (default depth 64) new events are dropped rather than stalling the agent, and each drop is counted and logged at debug level.
+
+| Event                 | Default | Emitted when                             |
+| --------------------- | ------- | ---------------------------------------- |
+| `device.added`        | yes     | a device is provisioned and registered   |
+| `device.removed`      | yes     | a device is deregistered                 |
+| `device.iface_opened` | yes     | a device's physical interface is opened  |
+| `device.iface_closed` | yes     | a device's physical interface is closed  |
+| `device.seen`         | no      | a device is marked seen (live heartbeat) |
+
+`device.seen` is **excluded by default**: the telemetry scheduler marks a device seen on every poll (milliseconds apart), so delivering it would flood the queue and crowd out the lifecycle events. Set `MG_AGENT_DEVICE_WEBHOOK_EVENTS` to an explicit comma-separated allowlist to change which events are delivered (e.g. `device.added,device.removed,device.seen`).
+
+Event payload:
+
+```json
+{
+  "event": "device.added",
+  "device_id": "63bdb473-02e6-457b-bb9a-773a18ab40a7",
+  "timestamp": "2026-06-23T08:00:00Z",
+  "device": { "id": "63bdb473-...", "name": "temp-sensor", "...": "..." }
+}
+```
+
+The full `device` record is included on `device.added`; other events carry the `device_id` only.
+
+### Authentication
+
+If `MG_AGENT_DEVICE_WEBHOOK_SECRET` is set, the agent signs the request body with HMAC-SHA256 and sends the signature in the `X-Agent-Webhook-Signature` header as `sha256=<hex>`. The raw secret never travels on the wire, and the receiver can verify both the sender and the payload integrity by recomputing the HMAC over the received body:
+
+```
+X-Agent-Webhook-Signature: sha256=9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08
+```
+
+Use an HTTPS endpoint for the webhook URL; the signature authenticates the payload but does not encrypt it.
+
 ## Configuration
 
 ### Environment Variables
 
-| Variable                    | Default | Description                                         |
-| --------------------------- | ------- | --------------------------------------------------- |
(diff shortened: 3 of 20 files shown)


api/transport.go
```diff
@@ -107,6 +107,18 @@ func MakeHandler(svc agent.Service, logger *slog.Logger, stream *logstream.Strea
 		EncodeResponse,
 		opts...,
 	).ServeHTTP)
+	r.Get("/devices/backup", kithttp.NewServer(
```
rodneyosodo, line 110: Update UI with these 2 endpoints too

docs/devices.md
```diff
@@ -57,16 +57,77 @@ When a device has a valid channel ID, the agent launches a background goroutine
...
-| `MG_AGENT_PROVISION_TOKEN`  |         | Token for provisioning API authentication           |
-| `MG_AGENT_RULES_ENGINE_URL` |         | URL for the Rules Engine (optional, for auto-rules) |
+| Variable                       | Default | Description                                                  |
+| ------------------------------ | ------- | ------------------------------------------------------------ |
+| `MG_AGENT_DEVICE_DB_PATH`      |         | Path to the BoltDB database file                             |
+| `MG_AGENT_PROVISION_URL`       |         | Base URL for Magistrala provisioning API                     |
+| `MG_AGENT_PROVISION_TOKEN`     |         | Token for provisioning API authentication                    |
+| `MG_AGENT_RULES_ENGINE_URL`    |         | URL for the Rules Engine (optional, for auto-rules)          |
+| `MG_AGENT_DEVICE_WEBHOOK_URL`  |         | Endpoint to receive device lifecycle events (empty disables) |
+| `MG_AGENT_DEVICE_WEBHOOK_SECRET` |       | Sent as `X-Agent-Webhook-Secret` on each webhook delivery    |
```
rodneyosodo: Align table

rodneyosodo (changes requested):

api/transport.go
JeffMboya: Done in 67dc2d5. Added **Backup** and **Restore** actions to the Devices page (`ui/src/pages/devices.tsx`):
- **Backup** → `GET /devices/backup`, downloads a timestamped `devices-backup-<ts>.json` (disabled when the registry is empty).
- **Restore** → file picker → validates the JSON, then a dialog to choose **Merge** (`?replace=false`) or **Replace** (`?replace=true`), shows the imported count, and reloads.

docs/devices.md
JeffMboya: Done in 67dc2d5. Realigned the environment-variable table — all column pipes now line up with the widest entry (`MG_AGENT_DEVICE_WEBHOOK_SECRET`).

rodneyosodo (commented):
.

rodneyosodo (changes requested):
A few issues worth addressing before merge, ordered by priority.

pkg/devicemgr/manager.go
```diff
@@ -107,7 +144,11 @@ func (m *Manager) Count() (int, error) {
 
 // MarkSeen updates LastSeen and Active for a device (called when a device sends data).
 func (m *Manager) MarkSeen(id string) error {
-	return m.store.MarkSeen(id)
+	if err := m.store.MarkSeen(id); err != nil {
+		return err
+	}
+	m.emit(EventDeviceSeen, id, nil)
```
rodneyosodo, line 151: This fires on every  call, but the scheduler calls  in a ~5ms poll loop (). That's up to 200  events/sec per device, which will overflow the 64-deep queue instantly and crowd out more important events. Could you debounce this or make  opt-in?

pkg/devicemgr/webhook.go
```diff
@@ -0,0 +1,155 @@
...
+// dropped rather than blocking the caller, since lifecycle notifications are
+// best-effort and must not stall device management.
+func (n *webhookNotifier) Notify(ev Event) {
+	if ev.Timestamp.IsZero() {
+		ev.Timestamp = time.Now().UTC()
+	}
+	select {
+	case n.events <- ev:
+	case <-n.done:
+	default:
```
rodneyosodo, line 166: Silent drop with no log or metric — operators won't know events are being lost. Maybe add a debug log or a dropped counter?

pkg/devicemgr/webhook.go
```diff
@@ -0,0 +1,155 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, n.cfg.URL, bytes.NewReader(body))
+	if err != nil {
+		return
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if n.cfg.Secret != "" {
+		req.Header.Set("X-Agent-Webhook-Secret", n.cfg.Secret)
+	}
+	resp, err := n.client.Do(req)
+	if err != nil {
```
rodneyosodo, line 255: A single transient network error permanently loses the event. Worth at least one retry for low-frequency events like /?

pkg/devicemgr/webhook.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), n.cfg.Timeout)
+	defer cancel()
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, n.cfg.URL, bytes.NewReader(body))
+	if err != nil {
+		return
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if n.cfg.Secret != "" {
+		req.Header.Set("X-Agent-Webhook-Secret", n.cfg.Secret)
```
rodneyosodo: Sending the secret as a cleartext header doesn't protect against payload tampering. Consider HMAC-signing the body instead. At minimum, worth a note in the docs that HTTPS is required.

api/transport.go
```diff
@@ -265,6 +277,16 @@ func decodeAddDeviceRequest(_ context.Context, r *http.Request) (any, error) {
 	return req, nil
 }
 
+func decodeRestoreDevicesRequest(_ context.Context, r *http.Request) (any, error) {
+	req := restoreDevicesReq{
+		Replace: strings.EqualFold(r.URL.Query().Get("replace"), "true"),
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req.Backup); err != nil {
+		return nil, mgerrors.Wrap(apiutil.ErrMalformedRequestBody, err)
+	}
```
rodneyosodo, line 291: No body size limit on the decoder — a large upload could exhaust memory. Maybe wrap with ?

api/requests.go
```diff
@@ -121,6 +124,11 @@ func (req addDeviceReq) validate() error {
 	return nil
 }
 
+type restoreDevicesReq struct {
```
rodneyosodo, line 127: Other request types implement . Any reason this one skips it? Even a nil-devices check here would be consistent.

pkg/devicemgr/backup.go
```diff
@@ -0,0 +1,124 @@
...
+func (m *Manager) Backup() (Backup, error) {
+	return m.store.Export()
+}
+
+// Restore loads devices from a Backup into the registry. When replace is true
+// the existing registry is cleared first; any open physical interfaces are
+// closed because their devices may no longer be present. It returns the number
+// of devices written.
+func (m *Manager) Restore(b Backup, replace bool) (int, error) {
+	if replace {
```
rodneyosodo, line 111: A concurrent  between  and  could open an interface for a device the replace is about to wipe. Probably fine in practice but worth a sentence in the docs.

pkg/devicemgr/manager.go
rodneyosodo, line 151: This fires on every `MarkSeen` call, but the scheduler calls `MarkSeen` in a ~5ms poll loop (`scheduler.go:179`). That's up to 200 `device.seen` events/sec per device — the 64-deep queue will overflow instantly and crowd out more important events. Could you debounce this or make `device.seen` opt-in?

pkg/devicemgr/webhook.go
rodneyosodo, line 166: Silent drop with no log or metric — operators won't know events are being lost. Maybe add a debug log or a dropped counter?

pkg/devicemgr/webhook.go
rodneyosodo, line 255: A single transient network error permanently loses the event. Worth at least one retry for low-frequency events like `device.added` / `device.removed`?

pkg/devicemgr/webhook.go
rodneyosodo: Sending the secret as a cleartext header doesn't protect against payload tampering. Consider HMAC-signing the body instead (e.g. `X-Agent-Webhook-Signature`). At minimum, worth noting in the docs that HTTPS is required.

api/transport.go
rodneyosodo, line 291: No body size limit on the decoder — a large upload could exhaust memory. Maybe wrap with `http.MaxBytesReader`?

api/requests.go
rodneyosodo, line 127: Other request types implement `validate()`. Any reason this one skips it? Even a nil-devices check here would be consistent with the rest of the codebase.

pkg/devicemgr/backup.go
rodneyosodo, line 111: A concurrent `OpenIface` between `closeAllIfaces()` and `store.Import()` could open an interface for a device the replace is about to wipe. Probably fine in practice but worth a sentence in the docs.

pkg/devicemgr/manager.go
JeffMboya: Fixed in c8c4a65. `device.seen` is now **excluded from the default event set** for exactly this reason — the scheduler poll loop would flood the queue. It is opt-in via `MG_AGENT_DEVICE_WEBHOOK_EVENTS` (comma-separated allowlist, e.g. `device.added,device.removed,device.seen`). The lifecycle test now asserts `device.seen` is not delivered by default, and `TestManager_WebhookEventOptIn` covers the opt-in path.

pkg/devicemgr/webhook.go
JeffMboya: Fixed in c8c4a65. Drops now increment an atomic counter (exposed via `Dropped()`) and log at debug level with the event type, device ID, and running total. Delivery failures (after retries, or non-retryable) also log at debug. The notifier takes a `Logger` through `WebhookConfig`, wired to the agent logger in `main.go`.

pkg/devicemgr/webhook.go
JeffMboya: Fixed in c8c4a65. `deliver()` now retries on network errors and `5xx` responses with a short backoff (default 2 retries, configurable via `WebhookConfig.Retries`); `4xx` is treated as permanent and not retried. The backoff aborts immediately on shutdown so `Close()` is not delayed. Added `TestManager_WebhookRetry` (first attempt 503, second 200).

pkg/devicemgr/webhook.go
JeffMboya: Fixed in c8c4a65. The raw secret is no longer sent. When a secret is set, the body is HMAC-SHA256 signed and sent as `X-Agent-Webhook-Signature: sha256=<hex>`, so the receiver verifies both authenticity and payload integrity by recomputing the HMAC over the received body. Docs now recommend an HTTPS endpoint. Added `TestManager_WebhookHMACSignature`.

api/transport.go
JeffMboya: Fixed in c8c4a65. `decodeRestoreDevicesRequest` now wraps the body in `io.LimitReader` at 16 MiB (`maxRestoreBodyBytes`) before decoding — ample for tens of thousands of device records, but bounded.

api/requests.go
JeffMboya: Fixed in c8c4a65. Added `restoreDevicesReq.validate()` (rejects any device with an empty ID, returning `400` instead of a downstream `500`) and the endpoint now calls it — matching the `addDeviceReq` convention.

pkg/devicemgr/backup.go
JeffMboya: Documented in c8c4a65. Added a note to both the `Restore` godoc and `docs/devices.md`: a replace restore is an administrative op, is not synchronized against concurrent `OpenIface`, and callers should quiesce device activity (or restart the agent) around it. Happy to add a coarse lock around close+import if you would prefer enforcement over a documented contract — let me know.

pkg/devicemgr/manager.go
JeffMboya: Addressed in c8c4a65: `device.seen` is excluded from the default event set and made opt-in via `MG_AGENT_DEVICE_WEBHOOK_EVENTS` (same fix as the parallel thread).

pkg/devicemgr/webhook.go
JeffMboya: Addressed in c8c4a65: drops now increment an atomic counter (`Dropped()`) and log at debug; delivery failures log too.

pkg/devicemgr/webhook.go
JeffMboya: Addressed in c8c4a65: `deliver()` retries on network errors / `5xx` with backoff (default 2, configurable); `4xx` is permanent.

pkg/devicemgr/webhook.go
JeffMboya: Addressed in c8c4a65: switched to HMAC-SHA256 body signing (`X-Agent-Webhook-Signature: sha256=<hex>`); raw secret is no longer sent. Docs recommend HTTPS.

api/transport.go
JeffMboya: Addressed in c8c4a65: restore body is capped at 16 MiB via `io.LimitReader` before decoding.

api/requests.go
JeffMboya: Addressed in c8c4a65: added `restoreDevicesReq.validate()` (empty device ID → `400`) and the endpoint calls it.

pkg/devicemgr/backup.go
JeffMboya: Addressed in c8c4a65: documented the replace-restore vs `OpenIface` race in the `Restore` godoc and docs; offered a coarse lock if you prefer enforcement.

rodneyosodo (approved):
LGTM

